PR #7, opened 2022-01-19
Sharp widget undock icon

luisrguerra:
Improved icon resolution. Remade icon aligned to 12x12 pixel grade for better resolution when the icon is resized to 12x12 at ui.
paintdotnet_69UZbgqoov

Current icon:
paintdotnet_al4aLs2UlM


ttlmax:
Thank you very much for the patch and the nice icon. I have pushed the patch as this [changeset](http://hg.savannah.gnu.org/hgweb/octave/rev/dfd2f1db0291) to the hg octave repository since the github mirror is not synchronized back to hg.

PR #11, opened 2022-02-14
applications-system and system-run new icons

luisrguerra:
The current gear icons are not looking good, I made new ones following the style used in the other Pull Requests I sent.

applications-system applications-system (1)

system-run system-run (1)

New:
preview-v2

Current:
old

New:
preview3

Current:
old2


luisrguerra:
I created a repository for these icons that I am developing and defined the license as Creative Commons Zero v1.0 Universal. To be in the public domain without restrictions.
[luisrguerra/cc0-fluent-icons](https://github.com/luisrguerra/cc0-fluent-icons)

siko1056:
Thank you very much for your great work on the icons and sharing your work :slightly_smiling_face:  :+1: 

Regarding `system-run`, I would personally confuse the gear icon with green arrow with an options menu, rather than a running my script button.  Again, just personal preference, but wouldn't a big green arrow suffice for intuition?  Maybe this is too similar to  the original? :roll_eyes:

luisrguerra:
> Regarding `system-run`, I would personally confuse the gear icon with green arrow with an options menu, rather than a running my script button. Again, just personal preference, but wouldn't a big green arrow suffice for intuition? Maybe this is too similar to the original? 🙄

A green play ▶️ icon  would would be more recognizable.

luisrguerra:
I changed the icons for a play symbol but kept the previous one renamed in case it is useful someday.
system-run
153879192-20ec8ced-7308-4b25-a86d-a54abaa278e9

luisrguerra:
I suggest replacing the gears for the file navigation menu with a folder with a gear symbol. This makes the menu easier to understand. It would be necessary to implement in the code
folder-settings

153878203-65454e1b-f019-437e-9d74-eda950dbb6e3

siko1056:
@luisrguerra Thanks again for the changes.  Looks good to me :+1: 

@ttlmax as the Octave GUI expert, how do you think about the new icons?  If you agree, I can add them via Mercurial.

siko1056:
According to #13, this will be handled once (almost) all icons are ready by @ttlmax.

ttlmax:
Yes, I will take care of all outstanding pull request regarding icons, when I have restructured the icon handling in the default branch.

ttlmax:
And changeset http://hg.savannah.gnu.org/hgweb/octave/rev/5dcdb03aa0e2 adds/updates the two icons from this pull request. If everything is okay, I will close this pull request

luisrguerra:
> And changeset http://hg.savannah.gnu.org/hgweb/octave/rev/5dcdb03aa0e2 adds/updates the two icons from this pull request. If everything is okay, I will close this pull request

### It remains to implement this icon (folder-settings), in case you are interested in doing this:
154473654-9a9e5d57-0050-4370-87c1-2d7f56297eba

ttlmax:
I really like this icon, but the problem is that "folder-settings" is not an icon following the [icon naming specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html). This leads to problems when using different icon themes, since other themes do not have this icon.
Having the octave theme as a fallback for the system or tango theme is only a n emergency solution since the style of the icons is really quite different.

ttlmax:
We could add a fallback icon as an additional argument for the function that is called for selecting the icon. When selecting "folder-settings", "applications-systems" has to be given as fallback.

luisrguerra:
> I really like this icon, but the problem is that "folder-settings" is not an icon following the [icon naming specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html). This leads to problems when using different icon themes, since other themes do not have this icon. Having the octave theme as a fallback for the system or tango theme is only a n emergency solution since the style of the icons is really quite different.

I even thought about this problem a while after I made the icons, because in Linux the folder icons are usually used as the system ones. And I looked in the Papirus theme icons as a reference and didn't find any name for the "folder-up" icon, for "folder-settings" there might be a similar one for settings folders but I'm not sure if it is present in most of Linux distros and would be suitable for that case.

luisrguerra:
I think that maybe the reason for not having these icons in the specs is because the function of these buttons are not common and maybe it is really necessary that they have their own icons to be better understood by the user

ttlmax:
I have now added two changesets where the first adds the possibility to provide a fall back icon with the icon request. This is then used for the folder-settings icon with applications-system as fallback. The second changeset sets go-up as fallback when folder-up is the desired icon.

Next, I will update the file license.md in the icon folder. What would be a suitable acknowledgement in this file for your work on most of the icons?

I am closing this pull request, since all new icons are now included in the octave icon theme. @luisrguerr, thank you very much for your icons.

luisrguerra:
> Next, I will update the file license.md in the icon folder. What would be a suitable acknowledgement in this file for your work on most of the icons?

OK, I'm building a repository [github.com/luisrguerra/cc0-fluent-icons](github.com/luisrguerra/cc0-fluent-icons) just for the icons to be better organized. When it's really complete (there are still some icons to be made) I think it would be interesting to fork to the GNU Octave organization on Github if you're interested.

> I am closing this pull request, since all new icons are now included in the octave icon theme. @luisrguerr, thank you very much for your icons.

It was nothing, anything you can contact me through my email luiseduardoribeiroguerra@gmail.com

PR #13, opened 2022-02-19
New save icons that are easier to understand and recognize

luisrguerra:
New save icons that are easier to understand and recognize

save-24pxsave-as-24px

preview2
Current:
153879215-6cf5c621-1e46-4fbd-ba9d-0391528b5492


ttlmax:
The new icons look nice! However, instead of exchanging one or two icons at a time, should we think about exchanging the complete icon set by a more modern one?

luisrguerra:
> The new icons look nice! However, instead of exchanging one or two icons at a time, should we think about exchanging the complete icon set by a more modern one?

Doing it gradually I believe it gets better done. You mean in making a single pull request? The problem with this is that some icons need code changes and others just need to accept the Pull Request. If you're talking about using a ready-made icon set (Example: Materia Design Icons), the most complete ones are monochromatic and I don't think they would look good in this program.

ttlmax:
I see you point.

Inspired by your work on the icons, I am thinking about providing different icon themes. In that case I would add your new icons to an "octave theme", which is a fallback for the special icons not included in the tango themes (breakpoints, widgets etc.). As long as we do not have new versions for all required icons, the tango theme will serve as fallback for the (default) octave-theme. 

If you agree, I will take care of this and your previous requests when I have re-structured the icon files (themes) in the default branch. What do you think?

Btw, I am not sure if any of the other developers has asked: The icons are created by yourself and are not from another free theme, where we would have to provide copyright notes and/or acknowledgements?

luisrguerra:
> Btw, I am not sure if any of the other developers has asked: The icons are created by yourself and are not from another free theme, where we would have to provide copyright notes and/or acknowledgements?

I made them from scratch, I'm also building a repository for it with the Creative Commons 0 license that gives which gives very broad permissions.
[luisrguerra/cc0-fluent-icons](https://github.com/luisrguerra/cc0-fluent-icons)

luisrguerra:
> Inspired by your work on the icons, I am thinking about providing different icon themes. In that case I would add your new icons to an "octave theme", which is a fallback for the special icons not included in the tango themes (breakpoints, widgets etc.). As long as we do not have new versions for all required icons, the tango theme will serve as fallback for the (default) octave-theme.

I believe support for different icon themes is something open source software projects should always consider.  Something that is also very relevant is to make the icons accessible in the software files to allow even a lay user to modify the appearance of the software.

luisrguerra:
When I have more time I will try to finish the rest of the icons and try to send them all together to be reviewed and implemented together.

luisrguerra:
I moved the changes from this Pull Request to [another](https://github.com/gnu-octave/octave/pull/14) with more content

PR #15, opened 2022-03-06
New icons part 2

luisrguerra:
Needs to be implemented in octave theme in code
view-zoom-in222 view-zoom-out32 view-zoom-original32 dialog-information32


ttlmax:
Pushed into mercurial repo with cset http://hg.savannah.gnu.org/hgweb/octave/rev/96a3c5548546

PR #16, opened 2022-03-07
New icons part 3

luisrguerra:


ttlmax:
Thanks for the improvement. Pushed with cset http://hg.savannah.gnu.org/hgweb/octave/rev/aee523d3036f

PR #18, opened 2022-03-19
New icons 5

luisrguerra:
go-home.png need to be implemented in the code.

I'm not very satisfied with the Home icon, after implementing it, in the future I will make a better version

preview 2


ttlmax:
Thanks for the nice icons, very much appreciated. I have pushed the patch with changeset http://hg.savannah.gnu.org/hgweb/octave/rev/a0e13472457d

PR #19, opened 2022-03-27
New icons 6

luisrguerra:
Preview:
preview 2

preview3v2

Octave 6.4:
old

Observation: The icons on "figure" window are rendered at 24x24 resolution but the icons are in 32x32, because of 32 is not a multiple of 24 the image became blurry even though the .svg image is align to 24x24 pixel grade.


ttlmax:
hanks for the icons. Merged into the default branch of the mercurial repo with changeset http://hg.savannah.gnu.org/hgweb/octave/rev/f14ad09af093

PR #20, opened 2022-09-12
add support for sparse computations in ode15{i,s} using Octave classes to be used when the SUNDIALS interface to KLU is not found ( GSoC '22)

sarrah-basta:


vivekd01:
The feature request is to add support for sparse computations in the ode15{i,s} solvers in Octave, specifically using Octave classes, as an alternative when the SUNDIALS interface to KLU is not found. This enhancement would improve the efficiency and performance of solving ordinary differential equations (ODEs) with sparse systems in Octave.

PR #22, opened 2024-02-13
Fix lower bound for GLP_DB

jo-37:
Found a small bug in the GLPK interface:
For doubly bound constraints (ctype "D"), the lower bound for `glp_set_row_bnds()` was not correct.

### DIFF
--- a/libinterp/dldfcn/__glpk__.cc
+++ b/libinterp/dldfcn/__glpk__.cc
@@ -167,7 +167,7 @@ glpk (int sense, int n, int m, double *c, int nz, int *rn, int *cn,
           break;
         }
 
-      glp_set_row_bnds (lp, i+1, typx, b[i], b[i]);
+      glp_set_row_bnds (lp, i+1, typx, typx == GLP_DB ? -b[i] : b[i], b[i]);
 
     }
 


mmuetzel:

Could you please show example code that currently does the wrong this and that would be fixed with this change?

It would be nice if you could add a built-in self test (BIST) that makes sure that this doesn't regress accidentally from a future change. `glpk.m` might be a good place for that BIST. E.g., similar to this existing one:
https://github.com/gnu-octave/octave/blob/ad9f517d08683f564ed7d1d8e85a0ae4f8c7236a/scripts/optimization/glpk.m#L656-L669

jo-37:
I am not a octave programmer/developer/user. Currently I'm trying to mimic Octave's interface to GLPK in Perl/PDL. For this purpose I extracted the function`glpk()` from `libinterp/dldfcn/__glpk__.cc`. Playing around with some parameters I spotted the GLP warning: `glp_simplex: row 1: lb = 1, ub = 1; incorrect bounds` when using a ctype of `'D'`. `msglev` must be nonzero to see it. An example task that will not be solved correctly:
```
sense = 1;
c = [-1, -1];
A = [1, 0; 0, 1];
b = [1,1];
ctype = ['D', 'D'];
lb = [-1, -1];
up = [1, 1];
```
The correct solution is:
```
xmin = [-1, -1]
fmin = 2
```
I just don't know how to make this into a BIST.

mmuetzel:
Thank you for providing the test case.
I could reproduce the issue (without your patch) and verify that your patch fixes it.

This is just a mirror of the main (Mercurial) repository. I'd like to push your change to the main repository. Could you please provide your (full) name and email address (that I could use as the author of this change)? 
If you are uncomfortable sharing that information, I could also make the change with myself as the author (but I don't want to appropriate your patch without your consent).

jo-37:
No problem with sharing my real name:
Jörg Sommrey <git@sommrey.de>

mmuetzel:
Thanks again for your contribution.
I added a BIST and pushed your change to the stable branch:
https://hg.savannah.gnu.org/hgweb/octave/rev/db32db01963a

It should be synchronized to the Git repository shortly.

PR #30, opened 2025-08-01
Add other libs

talregev:
Add other libs.
In vcpkg we compile octave dynamic with static libs. 
Example how it inside Qt5Core.pc file:

```
Libs: "-L${libdir}" -lQt5Core "-L${prefix}/lib" "-L${prefix}/lib/manual-link" -lpthread -lm ${prefix}/lib/libz.a -ldouble-conversion "-L${prefix}/lib/pkgconfig/../../lib" -lpcre2-16 ${prefix}/lib/libpng16.a ${prefix}/lib/libz.a
```

The direct linkage of zlib library is consider other lib, needed for compilation when compile the gui feature.

### DIFF
--- a/m4/acinclude.m4
+++ b/m4/acinclude.m4
@@ -981,7 +981,7 @@ AC_DEFUN([OCTAVE_CHECK_LIB], [
       if test -z "$m4_toupper(patsubst([$1], [-], [_]))_LDFLAGS"; then
         m4_toupper(patsubst([$1], [-], [_]))_LDFLAGS="$($PKG_CONFIG --libs-only-L m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
       fi
-      m4_toupper(patsubst([$1], [-], [_]))_LIBS="$($PKG_CONFIG --libs-only-l m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
+      m4_toupper(patsubst([$1], [-], [_]))_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
     ])
   fi
 
@@ -2280,7 +2280,7 @@ AC_DEFUN([OCTAVE_CHECK_QT_VERSION], [AC_MSG_CHECKING([Qt version $1])
       ## Retrieve Qt compilation and linker flags
       QT_CPPFLAGS="$($PKG_CONFIG --cflags-only-I $QT_MODULES | $SED -e 's/^ *$//')"
       QT_LDFLAGS="$($PKG_CONFIG --libs-only-L $QT_MODULES | $SED -e 's/^ *$//')"
-      QT_LIBS="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
+      QT_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other $QT_MODULES | $SED -e 's/^ *$//')"
 
       case $host_os in
         *darwin*)


mmuetzel:

If I understand correctly, some linkers (like `ld.bfd`) require that libraries that define symbols must appear after libraries that use these symbols.

Unless I'm misunderstanding the proposed change, the order of the flags would change from the order they appear in the .pc file. (Flags starting with `-l` would be sorted to the front.)
For the specific example with that package config file for Qt, that doesn't matter. But it could matter if it were a bit different.


Is there some way to get all flags that don't start with `-L` in the order that they appear? (I guess that is what we'd probably want.)

talregev:
The order stay the same, but without `--libs-only-other` the static flags are missing.

> Is there some way to get all flags that don't start with -L in the order that they appear? (I guess that is what we'd probably want.)

When you write them both flags: `--libs-only-l --libs-only-other` on a single command it keep the order of the flags.

mmuetzel:
Thanks for clarifying!

talregev:
@mmuetzel 
Take a look on another patch that we using in vcpkg. I didn't create it, so I cannot defend it:
https://github.com/microsoft/vcpkg/blob/master/ports/octave/run-mk-ops.diff

mmuetzel:
Took me a while to figure out when that could be needed. But if I understand correctly, that `awk`-script might be the first one to attempt to write to that folder in the build tree (before the compiler created any object files in that directory and implicitly created it). So, I think that is a reasonable change, too.

Since this is such a small change, I think we don't need to ask for permission to integrate it into the build system. If @dg0yt would prefer to be attributed with that change, I'd be happy to do that, too. In that case, we'd need a name and their email address.

dg0yt:
No credits required.

m4/acinclude.m4
```diff
@@ -981,7 +981,7 @@ AC_DEFUN([OCTAVE_CHECK_LIB], [
       if test -z "$m4_toupper(patsubst([$1], [-], [_]))_LDFLAGS"; then
         m4_toupper(patsubst([$1], [-], [_]))_LDFLAGS="$($PKG_CONFIG --libs-only-L m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
       fi
-      m4_toupper(patsubst([$1], [-], [_]))_LIBS="$($PKG_CONFIG --libs-only-l m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
+      m4_toupper(patsubst([$1], [-], [_]))_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
```
dg0yt, line 984: This is 
```suggestion
      m4_toupper(patsubst([$1], [-], [_]))_LIBS="$($PKG_CONFIG --libs m4_default([$9], [$1]) | $SED -e 's/^ *$//')"
```

m4/acinclude.m4
```diff
@@ -2280,7 +2280,7 @@ AC_DEFUN([OCTAVE_CHECK_QT_VERSION], [AC_MSG_CHECKING([Qt version $1])
       ## Retrieve Qt compilation and linker flags
       QT_CPPFLAGS="$($PKG_CONFIG --cflags-only-I $QT_MODULES | $SED -e 's/^ *$//')"
       QT_LDFLAGS="$($PKG_CONFIG --libs-only-L $QT_MODULES | $SED -e 's/^ *$//')"
-      QT_LIBS="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
+      QT_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other $QT_MODULES | $SED -e 's/^ *$//')"
```
dg0yt, line 2283:
```suggestion
      QT_LIBS="$($PKG_CONFIG --libs $QT_MODULES | $SED -e 's/^ *$//')"
```

mmuetzel:
Thinking about it: We might want to make sure that `$@-t.__DIR__` isn't an empty string before attempting to create a directory. It currently isn't afaict. But better save than sorry.
It shouldn't be too hard to check that though...

m4/acinclude.m4
mmuetzel: I don't think this is exactly the same. I think we still want the `-L` flags separately in `..._LDFLAGS`.

talregev: As I mention also in my PR, the `--libs` also bring the L flags.
upstream want to seperate it.

> Is there some way to get all flags that don't start with -L in the order that they appear?

dg0yt:
> When you write them both flags: `--libs-only-l --libs-only-other` on a single command it keep the order of the flags.

~As far as pkgconf is concerned, this is fully equivalent to `-libs`.~ 


> without --libs-only-other the static flags are missing.

There is some confusion here.

It is not related to static linkage at all. `--libs-only-l` lacks the directories (`-L`) and other flags, so the linker might not take the right link library (or evens fails to find a link library).

EDIT: Here, what is lacking from `--libs-only-l` is libs with absolute path.

In general, pkg-config must be invoked with `--static` for the static usage requirements. Assuming the pkgconfig files provide correct information, and that you don't mix static and shared libs.
(vcpkg doesn't need `--static` because it post-processes installed pc files for the actual linkage in the current build.)

> Is there some way to get all flags that don't start with `-L` in the order that they appear? (I guess that is what we'd probably want.)

To get a minimal list of ALL flags in the right order, you would need to add ALL modules to a single invocation of pkg-config / pkgconf. (pkgconf removes duplicate `-l` in a way which is compatible with traditional linkers.)

dg0yt:
I'm sorry. I was wrong about `--libs`. There is also `--libs-only-L`.

m4/acinclude.m4
dg0yt: My mistake.

talregev:
> As far as pkgconf is concerned, this is fully equivalent to `-libs`.
> 
> >

This is not true. Please verify with yourself before state that. 
`--libs-only-l --libs-only-other --libs-only-L` == `--libs`
`--libs-only-l --libs-only-other` this is bring all other flags without the L flags.

dg0yt:
Yeah, checked and rectified at the same time.

talregev:
Although the `--libs-only-other` is not connected the to static libs, the way we put the static links is with direct relative path to file itself, it consider as other libs, and it solve the problem that it adding the missing other libs files.

mmuetzel:
Afaict, `libtool` handles `LDFLAGS` and `LIBS` separately. So, we should at least attempt to separate them from the pkg-config files. I think that the change that @talregev proposes does a good-enough job at doing that.

If users want to override these flags, they could still do that with the corresponding configure flags (e.g., something along the lines of `QT_LIBS="$(pkgconf -static --libs-only-l --libs-only-other $QT_MODULES)" QT_LDFLAGS="$(pkgconf -static --libs-only-L $QT_MODULES)" QT_CPPFLAGS="$(pkgconf -static --cflags $QT_MODULES)"` while having set `QT_MODULES` to the needed Qt5 or Qt6 modules).

mmuetzel:
Some libraries need additional pre-processor definitions when linking to static libraries (like `-DSOMETHING_STATIC`).
I wonder if we should call `$PKG_CONFIG --cflags ...` instead of `$PKG_CONFIG --cflags-only-I ...` to accommodate that out of the box.
But I'd need to ask other devs if there is a good reason not to do that for everything. In any case, it's still possible to override all of that with configure flags...

mmuetzel:
I've pushed the changes from here to the upstream Mercurial repository. They should be appearing shortly here on this mirror.

I'll leave this PR open as a reminder for the other patch and potentially for the pre-processor flags from pkg-config files...

talregev:
@mmuetzel Thank you!

mmuetzel:
I've asked about `--cflags` vs. `--cflags-only-I` on the Octave Discourse forum:
https://octave.discourse.group/t/use-all-preprocessor-flags-from-pkg-config-files/6709

mmuetzel:
It looks like this change might have broken linking on macOS:
```
libtool: link: clang++ -std=gnu++17 -fPIC -Wall -W -Wshadow -Woverloaded-virtual -Wold-style-cast -Wformat -Wpointer-arith -Wwrite-strings -Wcast-align -Wcast-qual -fvisibility=hidden -O2 -g -o src/.libs/octave-gui src/octave_gui-main-gui.o src/octave_gui-octave-build-info.o -Wl,-bind_at_load  -L/opt/homebrew/opt/bison/lib -L/opt/homebrew/opt/gettext/lib -L/opt/homebrew/opt/icu4c/lib -L/opt/homebrew/opt/libiconv/lib -L/opt/homebrew/opt/readline/lib -L/opt/homebrew/opt/sqlite/lib -L/opt/homebrew/lib libgui/.libs/liboctgui.dylib -L/opt/homebrew/Cellar/hdf5/1.14.6/lib -L/opt/homebrew/opt/graphicsmagick/lib -L/opt/homebrew/Cellar/suite-sparse/7.11.0/lib -L/opt/homebrew/Cellar/fftw/3.3.10_2/lib -L/opt/homebrew/Cellar/arpack/3.9.1_1/lib -L/opt/homebrew/Cellar/pcre2/10.45/lib -L/opt/homebrew/opt/openblas/lib -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc/aarch64-apple-darwin23/15 -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc/aarch64-apple-darwin23/15/../../.. -L/opt/homebrew/opt/freetype/lib -L/usr/lib -L/opt/homebrew/opt/zstd/lib -L/opt/homebrew/opt/little-cms2/lib -L/opt/homebrew/opt/libpng/lib -L/opt/homebrew/opt/jpeg-turbo/lib -L/opt/homebrew/opt/webp/lib -L/opt/homebrew/opt/libtiff/lib -L/opt/homebrew/opt/jpeg-xl/lib -L/opt/homebrew/opt/libheif/lib -L/opt/homebrew/opt/jasper/lib -lqscintilla2_qt6 libmex/.libs/liboctmex.dylib /Users/runner/work/octave/octave/.build/libinterp/.libs/liboctinterp.dylib libinterp/.libs/liboctinterp.dylib /Users/runner/work/octave/octave/.build/liboctave/.libs/liboctave.dylib -lhdf5 /opt/homebrew/Cellar/graphicsmagick/1.3.45_1/lib/libGraphicsMagick++.dylib /opt/homebrew/Cellar/graphicsmagick/1.3.45_1/lib/libGraphicsMagick.dylib -llcms2 -lbz2 -lltdl -lz -lfontconfig -lfreetype -lgl2ps liboctave/.libs/liboctave.dylib -lcurl -lcholmod -lumfpack -lamd -lcamd -lcolamd -lccolamd -lcxsparse -lsuitesparseconfig -lspqr -larpack -lqrupdate -lfftw3_threads -lfftw3 -lfftw3f_threads -lfftw3f -lopenblas -lpcre2-8 -ldl -lemutls_w -lheapt_w -lgfortran -lquadmath libgnu/.libs/libgnu.a -lreadline -liconv -lpthread -lm -framework QtCore -framework QtCore5Compat -framework QtXml -framework QtGui -framework QtWidgets -framework QtPrintSupport -framework QtOpenGL -framework QtOpenGLWidgets -framework QtNetwork -framework QtSql -framework QtHelp -framework Carbon -framework OpenGL -pthread
ld: warning: -bind_at_load is deprecated on macOS
ld: framework 'QtCore' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The last successful run before this change used the following command to link that file:
```
libtool: link: clang++ -std=gnu++17 -fPIC -Wall -W -Wshadow -Woverloaded-virtual -Wold-style-cast -Wformat -Wpointer-arith -Wwrite-strings -Wcast-align -Wcast-qual -fvisibility=hidden -O2 -g -F/opt/homebrew/lib -o src/.libs/octave-gui src/octave_gui-main-gui.o src/octave_gui-octave-build-info.o -Wl,-bind_at_load  -L/opt/homebrew/Cellar/hdf5/1.14.6/lib -L/opt/homebrew/opt/graphicsmagick/lib -L/opt/homebrew/Cellar/suite-sparse/7.11.0/lib -L/opt/homebrew/Cellar/fftw/3.3.10_2/lib -L/opt/homebrew/opt/bison/lib -L/opt/homebrew/opt/gettext/lib -L/opt/homebrew/opt/icu4c/lib -L/opt/homebrew/opt/libiconv/lib -L/opt/homebrew/opt/readline/lib -L/opt/homebrew/opt/sqlite/lib -L/opt/homebrew/lib libgui/.libs/liboctgui.dylib -L/opt/homebrew/Cellar/arpack/3.9.1_1/lib -L/opt/homebrew/Cellar/pcre2/10.45/lib -L/opt/homebrew/opt/openblas/lib -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc/aarch64-apple-darwin23/15 -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc/aarch64-apple-darwin23/15/../../.. -L/opt/homebrew/opt/freetype/lib -L/usr/lib -L/opt/homebrew/opt/zstd/lib -L/opt/homebrew/opt/little-cms2/lib -L/opt/homebrew/opt/libpng/lib -L/opt/homebrew/opt/jpeg-turbo/lib -L/opt/homebrew/opt/webp/lib -L/opt/homebrew/opt/libtiff/lib -L/opt/homebrew/opt/jpeg-xl/lib -L/opt/homebrew/opt/libheif/lib -L/opt/homebrew/opt/jasper/lib libmex/.libs/liboctmex.dylib /Users/runner/work/octave/octave/.build/libinterp/.libs/liboctinterp.dylib libinterp/.libs/liboctinterp.dylib /Users/runner/work/octave/octave/.build/liboctave/.libs/liboctave.dylib liboctave/.libs/liboctave.dylib -lcurl -larpack -lqrupdate -lopenblas -lpcre2-8 -ldl -lemutls_w -lheapt_w -lgfortran -lquadmath libgnu/.libs/libgnu.a -lreadline -liconv -lqscintilla2_qt6 -lhdf5 /opt/homebrew/Cellar/graphicsmagick/1.3.45_1/lib/libGraphicsMagick++.dylib /opt/homebrew/Cellar/graphicsmagick/1.3.45_1/lib/libGraphicsMagick.dylib -llcms2 -lbz2 -lltdl -lz -lcholmod -lumfpack -lamd -lcamd -lcolamd -lccolamd -lcxsparse -lsuitesparseconfig -lspqr -lfftw3_threads -lfftw3 -lfftw3f_threads -lfftw3f -lfontconfig -lfreetype -lgl2ps -lpthread -lm -framework QtCore -framework QtCore5Compat -framework QtXml -framework QtGui -framework QtWidgets -framework QtPrintSupport -framework QtOpenGL -framework QtOpenGLWidgets -framework QtNetwork -framework QtSql -framework QtHelp -framework Carbon -framework OpenGL -pthread
```

The successful linker command included the flag `-F/opt/homebrew/lib` that I can no longer see in the new command.

mmuetzel:
That might be because `QT_LIBS` is now no longer empty when this is reached:
```
      case $host_os in
        *darwin*)
          ## Qt might be installed in framework
          if test -z "$QT_LIBS"; then
            QT_LDFLAGS="`$PKG_CONFIG --libs-only-other $QT_MODULES | tr ' ' '\n' | $GREP -e '-F' | uniq | tr '\n' ' '`"
            QT_LIBS="`$PKG_CONFIG --libs-only-other $QT_MODULES | tr ' ' '\n' | $GREP -v -e '-F' | uniq | tr '\n' ' '`"
          fi
        ;;
      esac
```

Edit: I think the intension for testing `test -z "$QT_LIBS"` was to not override configure flags that might have been set by the user.

talregev:
We can revert this change, and we only apply it in vcpkg.
Also I am suggesting to have a ci working when we do PRs to validate the solution also in here.
Let me know if you want me to create a PR that will trigger the ci when we do PR in here.

mmuetzel:
Running the CI here unfortunately interferes with the GitHub cache. Preferably, contributors would run the CI in their fork before opening a PR.

talregev:
> Running the CI here unfortunately interferes with the GitHub cache. Preferably, contributors would run the CI in their fork before opening a PR.

Why it interferes with the GitHub cache?
ASAICT the github PR don't touch the the GitHub cache, they using it, but not writing to it.
Also the PRs here never merged, just closed.

mmuetzel:
> ASAICT the github PR don't touch the the GitHub cache, they using it, but not writing to it.

See: 
https://github.com/gnu-octave/octave/blob/df420e385801937a46cb7de1cc8bee21422cb67e/.github/workflows/make-ubuntu.yaml#L154-L159

talregev:
I did the same on the GTSAM repo:
https://github.com/borglab/gtsam/blob/develop/.github/workflows/vcpkg.yml#L112

because they don't trigger this ci job on their main branch, the cache never saved, and each PR it start over.
It only saved during the PR itself.

mmuetzel:
> It only saved during the PR itself.

Correct. But it still adds to the overall cache limit. Which means that the caches for the "actual" CI runs (outside the PR) would be eliminated.

mmuetzel:
I've pushed a follow-up that should be fixing the regression on macOS:
https://hg.savannah.gnu.org/hgweb/octave/rev/91310231ca9e

talregev:
> I've pushed a follow-up that should be fixing the regression on macOS: https://hg.savannah.gnu.org/hgweb/octave/rev/91310231ca9e

Thank you!

talregev:
> I've pushed a follow-up that should be fixing the regression on macOS: https://hg.savannah.gnu.org/hgweb/octave/rev/91310231ca9e

The link is not accessible. It not find that reference.

mmuetzel:
> The link is not accessible. It not find that reference.

Savannah is having issues with the load recently (caused by data scrapers - likely LLM related). Please, try again. I pushed to the remote repository again. I'm seeing the commit now.

talregev:
> > The link is not accessible. It not find that reference.
> 
> Savannah is having issues with the load recently (caused by data scrapers - likely LLM related). Please, try again. I pushed to the remote repository again. I'm seeing the commit now.

Now I see your fix. Thank you!

mmuetzel:
Hmmm....
That didn't work.

dasergatskov:
It works for me if I explicitly specify `LDFLAGS="-F/opt/homebrew/opt/qt/lib"`
(not 100% sure but I think "-F" flag also provides path to include files, so you would not need all those "-I").

mmuetzel:
Before the changes from here:
```
    Qt CPPFLAGS:                   -I/opt/homebrew/lib/QtHelp.framework/Headers -I/opt/homebrew/lib/QtHelp.framework -I/opt/homebrew/lib/QtSql.framework/Headers -I/opt/homebrew/lib/QtSql.framework -I/opt/homebrew/lib/QtNetwork.framework/Headers -I/opt/homebrew/lib/QtNetwork.framework -I/opt/homebrew/lib/QtOpenGLWidgets.framework/Headers -I/opt/homebrew/lib/QtOpenGLWidgets.framework -I/opt/homebrew/lib/QtOpenGL.framework/Headers -I/opt/homebrew/lib/QtOpenGL.framework -I/opt/homebrew/lib/QtPrintSupport.framework/Headers -I/opt/homebrew/lib/QtPrintSupport.framework -I/opt/homebrew/lib/QtWidgets.framework/Headers -I/opt/homebrew/lib/QtWidgets.framework -I/opt/homebrew/lib/QtGui.framework/Headers -I/opt/homebrew/lib/QtGui.framework -I/opt/homebrew/lib/QtXml.framework/Headers -I/opt/homebrew/lib/QtXml.framework -I/opt/homebrew/lib/QtCore5Compat.framework/Headers -I/opt/homebrew/lib/QtCore5Compat.framework -I/opt/homebrew/lib/QtCore.framework/Headers -I/opt/homebrew/lib/QtCore.framework -I/opt/homebrew/share/qt/mkspecs/macx-clang -I/opt/homebrew/include
    Qt LDFLAGS:                    -F/opt/homebrew/lib 
    Qt GUI libraries:              -framework QtHelp -framework QtSql -framework QtNetwork -framework QtOpenGLWidgets -framework QtOpenGL -framework QtPrintSupport -framework QtWidgets -framework QtGui -framework QtXml -framework QtCore5Compat -framework QtCore  -lqscintilla2_qt6
```

After the changes from here:
```
    Qt CPPFLAGS:                   -I/opt/homebrew/lib/QtHelp.framework/Headers -I/opt/homebrew/lib/QtHelp.framework -I/opt/homebrew/lib/QtSql.framework/Headers -I/opt/homebrew/lib/QtSql.framework -I/opt/homebrew/lib/QtNetwork.framework/Headers -I/opt/homebrew/lib/QtNetwork.framework -I/opt/homebrew/lib/QtOpenGLWidgets.framework/Headers -I/opt/homebrew/lib/QtOpenGLWidgets.framework -I/opt/homebrew/lib/QtOpenGL.framework/Headers -I/opt/homebrew/lib/QtOpenGL.framework -I/opt/homebrew/lib/QtPrintSupport.framework/Headers -I/opt/homebrew/lib/QtPrintSupport.framework -I/opt/homebrew/lib/QtWidgets.framework/Headers -I/opt/homebrew/lib/QtWidgets.framework -I/opt/homebrew/lib/QtGui.framework/Headers -I/opt/homebrew/lib/QtGui.framework -I/opt/homebrew/lib/QtXml.framework/Headers -I/opt/homebrew/lib/QtXml.framework -I/opt/homebrew/lib/QtCore5Compat.framework/Headers -I/opt/homebrew/lib/QtCore5Compat.framework -I/opt/homebrew/lib/QtCore.framework/Headers -I/opt/homebrew/lib/QtCore.framework -I/opt/homebrew/share/qt/mkspecs/macx-clang -I/opt/homebrew/include
    Qt LDFLAGS:                    
    Qt GUI libraries:              -F/opt/homebrew/lib -framework QtHelp -framework QtSql -framework QtNetwork -framework QtOpenGLWidgets -framework QtOpenGL -framework QtPrintSupport -framework QtWidgets -framework QtGui -framework QtXml -framework QtCore5Compat -framework QtCore -lqscintilla2_qt6
```

So, we'd probably need to move the `-F` flag back from `QT_LIBS` to `QT_LDFLAGS`.

I thought that would happen with the following:
```
    if test $build_qt_gui = yes; then
      ## Retrieve Qt compilation and linker flags
      ac_octave_save_QT_LIBS="$QT_LIBS"
      QT_CPPFLAGS="$($PKG_CONFIG --cflags-only-I $QT_MODULES | $SED -e 's/^ *$//')"
      QT_LDFLAGS="$($PKG_CONFIG --libs-only-L $QT_MODULES | $SED -e 's/^ *$//')"
      QT_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other $QT_MODULES | $SED -e 's/^ *$//')"

      case $host_os in
        *darwin*)
          ## Qt might be installed in framework
          ac_octave_QT_LIBS_only_l="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
          if test -z "$ac_octave_save_QT_LIBS" && test -z "$ac_octave_QT_LIBS_only_l"; then
            QT_LDFLAGS="`$PKG_CONFIG --libs-only-other $QT_MODULES | tr ' ' '\n' | $GREP -e '-F' | uniq | tr '\n' ' '`"
            QT_LIBS="`$PKG_CONFIG --libs-only-other $QT_MODULES | tr ' ' '\n' | $GREP -v -e '-F' | uniq | tr '\n' ' '`"
          fi
        ;;
      esac
    fi
```

`QT_LIBS` is empty initially (if the user didn't explicitly set a value for it in the configure flags). We are storing that (empty) value in `ac_octave_save_QT_LIBS`. Then, we query the flags from the pkg-config files of Qt. Since we've added  `--libs-only-other` to the pkg-config flags, `QT_LIBS` is no longer empty after that step.

Now comes the macOS specific part:
In older versions of Qt, their pkg-config files returned `-L` and `-l` flags. So, we are done.
In newer versions of Qt, they no longer return `-l` flags. So, we probe for that with `--libs-only-l`.
If the user didn't supply flags for `QT_LIBS` initially (i.e., if `ac_octave_save_QT_LIBS` is empty) and if the pkg-config files for Qt didn't return any `-l` flags (i.e., if `ac_octave_QT_LIBS_only_l` is empty), we try to separate out the `-F` flag from the remainder.

I don't know why that doesn't work.

@dasergatskov: Could you please run the following commands on macOS (do not set `QT_LIBS` before that test)?
```
QT_MODULES="Qt6Core Qt6Gui Qt6Help Qt6Network Qt6OpenGL Qt6OpenGLWidgets Qt6PrintSupport Qt6Widgets Qt6Xml Qt6Core5Compat"
pkg-config --libs-only-l --libs-only-other $QT_MODULES
ac_octave_QT_LIBS_only_l="$(pkg-config --libs-only-l $QT_MODULES | sed -e 's/^ *$//')"
ac_octave_save_QT_LIBS="$QT_LIBS"
echo $ac_octave_QT_LIBS_only_l
echo $ac_octave_save_QT_LIBS
if test -z "$ac_octave_save_QT_LIBS" && test -z "$ac_octave_QT_LIBS_only_l"; then
  echo "both are empty"
fi
```

dasergatskov:
```
 % QT_MODULES="Qt6Core Qt6Gui Qt6Help Qt6Network Qt6OpenGL Qt6OpenGLWidgets Qt6PrintSupport Qt6Widgets Qt6Xml Qt6Core5Compat"
pkg-config --libs-only-l --libs-only-other $QT_MODULES
ac_octave_QT_LIBS_only_l="$(pkg-config --libs-only-l $QT_MODULES | sed -e 's/^ *$//')"
ac_octave_save_QT_LIBS="$QT_LIBS"
echo $ac_octave_QT_LIBS_only_l
echo $ac_octave_save_QT_LIBS
if test -z "$ac_octave_save_QT_LIBS" && test -z "$ac_octave_QT_LIBS_only_l"; then
  echo "both are empty"
fi

-F/opt/homebrew/lib -framework QtHelp -framework QtSql -framework QtNetwork -framework QtOpenGLWidgets -framework QtOpenGL -framework QtPrintSupport -framework QtWidgets -framework QtGui -framework QtXml -framework QtCore5Compat -framework QtCore


both are empty
```

mmuetzel:
Hmmm...
That is what I expected. I wonder why it doesn't do that in the CI.

dasergatskov:
I did have this linker problem until set the "-F" flag.

dasergatskov:
I tried again w/o this flag:
```
/bin/sh ./libtool  --tag=CXX   --mode=link clang++ -std=gnu++17 -fPIC -pthread -Wall -W -Wshadow -Woverloaded-virtual -Wold-style-cast -Wformat -Wpointer-arith -Wwrite-strings -Wcast-align -Wcast-qual -fvisibility=hidden -g3 -O3 -march=native -flto=thin -Wno-deprecated-declarations      -L/opt/homebrew/lib -L/opt/homebrew/opt/libiconv/lib -L/opt/homebrew/opt/readline/lib -L/usr/local/lib -o src/octave-gui src/octave_gui-main-gui.o src/octave_gui-octave-build-info.o libgui/liboctgui.la libmex/liboctmex.la libinterp/liboctinterp.la liboctave/liboctave.la libgnu/libgnu.la  -lpthread -lm -ljemalloc
libtool: link: clang++ -std=gnu++17 -fPIC -Wall -W -Wshadow -Woverloaded-virtual -Wold-style-cast -Wformat -Wpointer-arith -Wwrite-strings -Wcast-align -Wcast-qual -fvisibility=hidden -g3 -O3 -march=native -flto=thin -Wno-deprecated-declarations -o src/.libs/octave-gui src/octave_gui-main-gui.o src/octave_gui-octave-build-info.o -Wl,-bind_at_load  -L/opt/homebrew/lib -L/opt/homebrew/opt/libiconv/lib -L/opt/homebrew/opt/readline/lib -L/usr/local/lib libgui/.libs/liboctgui.dylib -L/opt/homebrew/Cellar/hdf5/1.14.6/lib -L/opt/homebrew/opt/graphicsmagick/lib -L/opt/homebrew/Cellar/suite-sparse/7.11.0/lib -L/opt/homebrew/Cellar/fftw/3.3.10_2/lib -L/opt/homebrew/Cellar/arpack/3.9.1_1/lib -L/opt/homebrew/Cellar/pcre2/10.45/lib -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc/aarch64-apple-darwin24/15 -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc -L/opt/homebrew/Cellar/gcc/15.1.0/bin/../lib/gcc/current/gcc/aarch64-apple-darwin24/15/../../.. -L/opt/homebrew/opt/freetype/lib -L/usr/lib -L/opt/homebrew/opt/zstd/lib -L/opt/homebrew/opt/little-cms2/lib -L/opt/homebrew/opt/libpng/lib -L/opt/homebrew/opt/jpeg-turbo/lib -L/opt/homebrew/opt/webp/lib -L/opt/homebrew/opt/libtiff/lib -L/opt/homebrew/opt/jpeg-xl/lib -L/opt/homebrew/opt/libheif/lib -L/opt/homebrew/opt/jasper/lib -lqscintilla2_qt6 libmex/.libs/liboctmex.dylib /Users/dmitri/src/octave/out_gopt/libinterp/.libs/liboctinterp.dylib libinterp/.libs/liboctinterp.dylib /Users/dmitri/src/octave/out_gopt/liboctave/.libs/liboctave.dylib -lhdf5 /opt/homebrew/Cellar/graphicsmagick/1.3.45_1/lib/libGraphicsMagick++.dylib /opt/homebrew/Cellar/graphicsmagick/1.3.45_1/lib/libGraphicsMagick.dylib -llcms2 -lbz2 -lltdl -lz -lfontconfig -lfreetype -lX11 -lgl2ps liboctave/.libs/liboctave.dylib -lcurl -lcholmod -lumfpack -lamd -lcamd -lcolamd -lccolamd -lcxsparse -lsuitesparseconfig -lspqr -larpack -lqrupdate -lfftw3_threads -lfftw3 -lfftw3f_threads -lfftw3f -lflexiblas -lpcre2-8 -ldl -lemutls_w -lheapt_w -lgfortran -lquadmath libgnu/.libs/libgnu.a -lreadline -liconv -lpthread -lm -ljemalloc -framework QtCore -framework QtCore5Compat -framework QtXml -framework QtGui -framework QtWidgets -framework QtPrintSupport -framework QtOpenGL -framework QtOpenGLWidgets -framework QtNetwork -framework QtSql -framework QtHelp -framework Carbon -framework OpenGL -pthread
ld: warning: -bind_at_load is deprecated on macOS
ld: framework 'QtCore' not found
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [Makefile:16532: src/octave-gui] Error 1
gmake[2]: Leaving directory '/Users/dmitri/src/octave/out_gopt'
gmake[1]: *** [Makefile:28720: all-recursive] Error 1
gmake[1]: Leaving directory '/Users/dmitri/src/octave/out_gopt'
gmake: *** [Makefile:11763: all] Error 2
```
I do not see "-F" flag here. But in the configure I see it:
```
  QRUPDATE libraries:            -lqrupdate
  Qt CPPFLAGS:                   -I/opt/homebrew/lib/QtHelp.framework/Headers -I/opt/homebrew/lib/QtHelp.framework -I/opt/homebrew/lib/QtSql.framework/Headers -I/opt/homebrew/lib/QtSql.framework -I/opt/homebrew/lib/QtNetwork.framework/Headers -I/opt/homebrew/lib/QtNetwork.framework -I/opt/homebrew/lib/QtOpenGLWidgets.framework/Headers -I/opt/homebrew/lib/QtOpenGLWidgets.framework -I/opt/homebrew/lib/QtOpenGL.framework/Headers -I/opt/homebrew/lib/QtOpenGL.framework -I/opt/homebrew/lib/QtPrintSupport.framework/Headers -I/opt/homebrew/lib/QtPrintSupport.framework -I/opt/homebrew/lib/QtWidgets.framework/Headers -I/opt/homebrew/lib/QtWidgets.framework -I/opt/homebrew/lib/QtGui.framework/Headers -I/opt/homebrew/lib/QtGui.framework -I/opt/homebrew/lib/QtXml.framework/Headers -I/opt/homebrew/lib/QtXml.framework -I/opt/homebrew/lib/QtCore5Compat.framework/Headers -I/opt/homebrew/lib/QtCore5Compat.framework -I/opt/homebrew/lib/QtCore.framework/Headers -I/opt/homebrew/lib/QtCore.framework -I/opt/homebrew/share/qt/mkspecs/macx-clang -I/opt/homebrew/include
  Qt LDFLAGS:                    
  Qt GUI libraries:              -F/opt/homebrew/lib -framework QtHelp -framework QtSql -framework QtNetwork -framework QtOpenGLWidgets -framework QtOpenGL -framework QtPrintSupport -framework QtWidgets -framework QtGui -framework QtXml -framework QtCore5Compat -framework QtCore -lqscintilla2_qt6
```

mmuetzel:
Please, try again after an update. The -F flag should be in Qt LDFLAGS again now.

I was looking into that at some point in the past. If I recall correctly, `libtool` is pretty picky when it comes to which flags it lets through. It allows `-F` only in `LDFLAGS` but not in `LIBS`.

dasergatskov:
Yes, it is fine now (4abf3eb9171a)

mmuetzel:
@dg0yt: Do you recall why you needed https://github.com/microsoft/vcpkg/blob/master/ports/octave/run-mk-ops.diff ?

I suspected that the `liboctave/operators` folder in the build tree might be created too late. But for me, it is already created while running the configure script. So, there is nothing for the Makefile to do at that point.
Which issue did you run into without that patch?

talregev:
Thank you for your fix!

dg0yt:
> I wonder if we should call `$PKG_CONFIG --cflags ...` instead of `$PKG_CONFIG --cflags-only-I ...` to accommodate that out of the box.

> I was looking into that at some point in the past. If I recall correctly, `libtool` is pretty picky when it comes to which flags it lets through. It allows `-F` only in `LDFLAGS` but not in `LIBS`.

libtool pick with `-F`? IDK, but the standard solution for processing pkg-config information in autoconf, `pkg_check_modules`, only does `--clags` and `--libs` (yielding `FOO_CFLAGS` and `FOO_LIBS`). This is good enough for virtually every case I encountered in vcpkg ports. But admittedly vpckg does avoid building frameworks.

(Coincidentally I immediately wondered about frameworks when the `--libs-only-L` came up today.)

dg0yt:
> Do you recall why you needed https://github.com/microsoft/vcpkg/blob/master/ports/octave/run-mk-ops.diff ?

I can't remember the details. Most like a race in the massive parallel build in CI, or some other side effect of the port or CI infrastructure.
We could run a build in the CI with the patch removed. The vcpkg owners are usually happy about patches being removed.

mmuetzel:
> We could run a build in the CI with the patch removed. The vcpkg owners are usually happy about patches being removed.

Thanks. I would appreciate if we could understand why the patch is needed before accepting.

talregev:
@mmuetzel 
@dg0yt Tested without this patch. We are going to remove it.

dg0yt:
@mmuetzel I'm still looking at it independently.

mmuetzel:
Which configure flags are you using?
While testing locally a plain configure (without flags) created that folder in the build tree.
Maybe, there are some (combinations of?) flags that cause the folder to not be created while running the configure script.

dg0yt:
I can't reproduce it now, even with the octave version where the patch was added. However, there isn't the same autotools environment now, and maybe also not the same flags. AFAIU the folder is created during configure to host the `.Plo` files in the `.deps` subdir. I assume this it is part of the step logged as `config.status: executing depfiles commands`. But I have no idea which conditions apply to the actual commands in this action.

mmuetzel:
Good hint.

Automake allows to disable dependency tracking by configuring with `--disable-dependency-tracking`. I usually never use that option because it causes issues with incremental builds. But if you are building only once, using that option is probably ok.
When configuring with that option, the `liboctave/operators` folder is *not* created in the build tree by the configure script. I can see how that could be causing issues when trying to generate files in a folder that doesn't yet exist.

I pushed the following change that should be creating the corresponding folders when the output of generator scripts is redirected to some file in the build tree:
https://hg.savannah.gnu.org/hgweb/octave/rev/2616f35310cb

I hope I didn't miss some generators. But the one that was important for you is included in that changeset.

I think this was the last change for issues about which we've talked here.
All of the changes that have been motivated by this PR should be part of Octave 11 when it will be released.

I'll close this PR after a successful round of CI (buildbots and here).

dg0yt:
> Automake allows to disable dependency tracking by configuring with `--disable-dependency-tracking`. I usually never use that option because it causes issues with incremental builds. But if you are building only once, using that option is probably ok.

Packaging is non-incremental, and the vcpkg maintainer functions for make-based builds indeed use `--disable-dependency-tracking`: On MSYS2/Windows, the effect of saving extra processes and forks is significant. (Octave doesn't build with MSVC, but it uses the same maintainer functions.)

talregev:
> I'll close this PR after a successful round of CI (buildbots and here).

Why not let the CI run, then push the commit? It save much more time like that.

mmuetzel:
> Why not let the CI run, then push the commit? It save much more time like that.

That isn't possible and it isn't true either.
The buildbots take approx. 24 hours *after* a change was pushed to complete their tasks.
Waiting for parts of the CI *before* pushing wouldn't save any time. It would take longer.

mmuetzel:
> Packaging is non-incremental, and the vcpkg maintainer functions for make-based builds indeed use --disable-dependency-tracking: On MSYS2/Windows, the effect of saving extra processes and forks is significant.


Good to know. Maybe, we could use that flag for the non-incremental builds in the CI (at least on Windows), too.
I"d need to experiment with that. At one time, the build stopped locally after only a small part was done locally for me with that flag. So, there might still be issues that need fixing before we can do that reliably. Or I did something wrong...

talregev:
> > Why not let the CI run, then push the commit? It save much more time like that.
> 
> That isn't possible and it isn't true either. The buildbots take approx. 24 hours _after_ a change was pushed to complete their tasks. Waiting for parts of the CI _before_ pushing wouldn't save any time. It would take longer.

I mean running the gihub ci before push your commit to the public repo.
It doesn't take 24 hours, and you also validate your commits with that.
Take example from my future PR:
https://github.com/talregev/octave/pull/4

mmuetzel:
> I mean running the gihub ci before push your commit to the public repo.

I don't know how many times I have to repeat myself. And I'm only answering because I try to be friendly.

We don't want to trigger the GitHub workflows on PRs in this repository.

Also, this is just a read-only mirror. So, opening PRs here and wait for (incomplete) CI results just to close the PR and do the same changes again on the main Mercurial repository would slow the development process down a lot

> It doesn't take 24 hours, and you also validate your commits with that.

Obviously you don't know that the CI of Octave includes more than just the GitHub workflows in this read-only mirror. Or you wouldn't make such bold claims. When I wrote "buildbots", I meant buildbots. E.g., these:
https://buildbot.octave.org/

And there are more automatic tests depending on which branch was changed...

talregev:
I didn't trigger the CI in here, it on my fork repo. 
More detail answer on my PR.

mmuetzel:
No further issues after a round of automated tests.

Closing this PR.

PR #31, opened 2025-08-01
debug osx

talregev:
Add ci for debug.

### DIFF
--- a/.github/workflows/make-macos.yaml
+++ b/.github/workflows/make-macos.yaml
@@ -2,6 +2,7 @@ name: make-macos
 on:
   workflow_dispatch:
   push:
+  pull_request:
 concurrency: ci-macos-${{ github.ref }}
 
 jobs:
--- a/m4/acinclude.m4
+++ b/m4/acinclude.m4
@@ -2278,7 +2278,6 @@ AC_DEFUN([OCTAVE_CHECK_QT_VERSION], [AC_MSG_CHECKING([Qt version $1])
 
     if test $build_qt_gui = yes; then
       ## Retrieve Qt compilation and linker flags
-      ac_octave_save_QT_LIBS="$QT_LIBS"
       QT_CPPFLAGS="$($PKG_CONFIG --cflags-only-I $QT_MODULES | $SED -e 's/^ *$//')"
       QT_LDFLAGS="$($PKG_CONFIG --libs-only-L $QT_MODULES | $SED -e 's/^ *$//')"
       QT_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other $QT_MODULES | $SED -e 's/^ *$//')"
@@ -2286,8 +2285,8 @@ AC_DEFUN([OCTAVE_CHECK_QT_VERSION], [AC_MSG_CHECKING([Qt version $1])
       case $host_os in
         *darwin*)
           ## Qt might be installed in framework
-          ac_octave_QT_LIBS_only_l="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
-          if test -z "$ac_octave_save_QT_LIBS" && test -z "$ac_octave_QT_LIBS_only_l"; then
+          QT_LIBS="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
+          if test -z "$QT_LIBS"; then
             QT_LDFLAGS="`$PKG_CONFIG --libs-only-other $QT_MODULES | tr ' ' '\n' | $GREP -e '-F' | uniq | tr '\n' ' '`"
             QT_LIBS="`$PKG_CONFIG --libs-only-other $QT_MODULES | tr ' ' '\n' | $GREP -v -e '-F' | uniq | tr '\n' ' '`"
           fi


m4/acinclude.m4
```diff
@@ -2278,16 +2278,15 @@ AC_DEFUN([OCTAVE_CHECK_QT_VERSION], [AC_MSG_CHECKING([Qt version $1])
 
     if test $build_qt_gui = yes; then
       ## Retrieve Qt compilation and linker flags
-      ac_octave_save_QT_LIBS="$QT_LIBS"
```
talregev, line 2281: QT_LIBS will overwrite anyway even if the user set it.

m4/acinclude.m4
```diff
@@ -2278,16 +2278,15 @@ AC_DEFUN([OCTAVE_CHECK_QT_VERSION], [AC_MSG_CHECKING([Qt version $1])
...
       QT_CPPFLAGS="$($PKG_CONFIG --cflags-only-I $QT_MODULES | $SED -e 's/^ *$//')"
       QT_LDFLAGS="$($PKG_CONFIG --libs-only-L $QT_MODULES | $SED -e 's/^ *$//')"
       QT_LIBS="$($PKG_CONFIG --libs-only-l --libs-only-other $QT_MODULES | $SED -e 's/^ *$//')"
 
       case $host_os in
         *darwin*)
           ## Qt might be installed in framework
-          ac_octave_QT_LIBS_only_l="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
-          if test -z "$ac_octave_save_QT_LIBS" && test -z "$ac_octave_QT_LIBS_only_l"; then
+          QT_LIBS="$($PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//')"
```
talregev, line 2288: This is not empty on osx.

m4/acinclude.m4
mmuetzel: There is the following condition earlier:
```
  if test $build_qt_gui = yes \
      && test -n "$QT_CPPFLAGS" \
      && test -n "$QT_LIBS"; then
```
which ends up in a different branch. And we need to save whether the user initially didn't set anything...

m4/acinclude.m4
mmuetzel: If it is not empty, then we are fine. That would mean that we don't deal with frameworks but with "normal" libraries.

.github/workflows/make-macos.yaml
```diff
@@ -2,6 +2,7 @@ name: make-macos
 on:
   workflow_dispatch:
   push:
+  pull_request:
```
mmuetzel, line 5: Why did you enable the CI on PRs even if I explicitly told you not to do that?

mmuetzel: Enable actions on your fork instead.

m4/acinclude.m4
talregev: Do you want to add that fix?
I am not sure my fix is correct, that why I try it on the ci.

talregev: I am waiting for the ci.

mmuetzel: No. That is already there...

m4/acinclude.m4
talregev: I am not sure of that, I just assume that I didn't see the -F flags before my change.

.github/workflows/make-macos.yaml
talregev: ok.

m4/acinclude.m4
mmuetzel: But you are right. QT_LIBS was already overridden in that branch before the changes from #30. So, we could probably continue to do that and ignore whatever the user might have set.

talregev: I close this PR because I want to to test it first before announced the fix. 
I need the CI, and I not allowed it here. so I open a PR on my fork instead.

talregev: When the fix will be ready, I will open a new PR without CI.

m4/acinclude.m4
mmuetzel: You should only see a `-F` flag if the other flags are `-f`. Or the other way round: If there are `-l` flags in the linker flags for Qt, there won't be a `-F` flag.

talregev: Take a look:
https://github.com/gnu-octave/octave/actions/runs/16677884718/job/47209059289#step:10:10056

m4/acinclude.m4
mmuetzel: You don't need to open a PR in your fork. Just push to a branch on your fork and let the CI run there.

talregev: I am working differently from you. It really easy for me to work with PR. and it really hard for me without it, that why I do it.

talregev: In the PR I can see all the changes, all the commits and all the CI jobs in one place. To take a look for different places these, it hard for my mind, because I get used to work with PRs.
I hope it explain a little bit.

m4/acinclude.m4
mmuetzel: Ok. So `$PKG_CONFIG --libs-only-l $QT_MODULES | $SED -e 's/^ *$//'` *is* empty on macOS. (I don't know where you are testing on OS X. Do you have access to an old MacBook?)

talregev: I testing only in the ci.
I didn't see the values as I say (that why I say I am not sure) I just see the outputs of the flags as I show you in the link.

talregev: If your fix is good, that fine. If not, I have a fix that will fix it after testing in the CI.

m4/acinclude.m4
mmuetzel: You don't need to open a PR in your fork. But of course you can.

In any case, you shouldn't enable the CI on PRs because the CI will already trigger on a push to your fork anyway.

Did you know that you can selectively disable actions on your fork using the "three dots" menu in the upper right corner after selecting a workflow on the "Actions" panel? That way you can have the CI being triggered only for the workflows you are interested in currently.

Almost all workflows have a workflow_dispatch trigger, too. So, you can trigger them manually on your fork even if you had it disabled while you pushed to your branch.

talregev: I already know I can selective workflow on my Actions. I already know the job is running when I do the push.
But again, because we work differently, and you never try to work as I do, it will hard for you to understand why I do insist to open a PR.

mmuetzel: I understand how opening PRs is useful. And I never told you not to do that.
The only requirement here would be: Do not trigger workflows on pull requests in this mirror.

m4/acinclude.m4
talregev: ok, the CI is green. no need for my fix. Thank you for your fix!

m4/acinclude.m4
talregev: ok that a fair request, and I sorry I trigger the CI in here. I will not do it again.

PR #40, opened 2026-01-22
median: use nth_element for linear-time selection on dense inputs

beingamanforever:
As pointed out in the [community](https://octave.discourse.group/t/in-octave-median-is-o-n-log-n-no-asymptotic-gain-over-sort-extra-hidden-work/7276) about the hidden cost of median(x) and possible fix.
### Summary

This change replaces the full sort used in `median()` for dense inputs with order-statistic selection via `nth_element`, reducing average complexity from **O(n log n)** to **O(n)**. Sparse inputs retain the original sort-based code path to preserve historical behavior and output sparsity.

---

### Algorithmic change

Previous implementation:
- `sort(x)` followed by indexing middle element(s)

New implementation:
- `nth_element(x, k)` for odd length
- `nth_element(x, [k, k+1])` for even length

This computes only the required order statistics without fully sorting.

---

### NaN handling

NaN semantics are preserved explicitly:

- `"omitnan"`: NaNs are filtered before selection.
- `"includenan"`: slices containing NaNs return NaN.

This matches the behavior previously provided implicitly by `sort()`

---
### Sparse behavior

Sparse inputs continue to use the original sort-based code path to preserve:
- sparsity of outputs,
- historical indexing behavior,
- and memory characteristics.

Dense inputs use the new `nth_element` path.

---
I went a step ahead and ran a benchmark to see if it's really doing the speed up as promised, here are the results below. I am also attaching the benchmarking script.

### Performance

Benchmarks on large dense inputs (Octave 9.x, Apple clang):

| Case | Size | Before (s) | After (s) | Speedup |
|------|------|------------|-----------|---------|
| Vector | 10,000 | 0.001555 | 0.000634 | 2.5× |
| Vector | 100,000 | 0.008014 | 0.001303 | 6.2× |
| Vector | 500,000 | 0.044739 | 0.005207 | 8.6× |
| Vector | 1,000,000 | 0.095003 | 0.012283 | 7.7× |
| Matrix (n×100) | 10,000 | 0.060745 | 0.015934 | 3.8× |
| Matrix (n×100) | 100,000 | 0.754142 | 0.126997 | 5.9× |
| Matrix (n×100) | 500,000 | 4.648894 | 0.590293 | 7.9× |
| Matrix (n×100) | 1,000,000 | 9.967157 | 1.471443 | 6.8× |
| omitnan (n×100) | 10,000 | 0.071202 | 0.015734 | 4.5× |
| omitnan (n×100) | 100,000 | 0.717203 | 0.117018 | 6.1× |
| omitnan (n×100) | 500,000 | 4.470758 | 0.560758 | 8.0× |
| omitnan (n×100) | 1,000,000 | 9.405867 | 1.238683 | 7.6× |

Speedup grows with input size, consistent with removal of the `log n` factor.

---

bench_median_comparison

Benchmarking Script can be found [here](https://pastebin.com/7iB8wjY4)
CSV results from which plots were generated can be found [here](https://github.com/user-attachments/files/24808338/bench_results.csv)

---
## Notes
- No semantic or API changes.
- Only the internal selection algorithm is modified.
- All existing tests pass
- All existing BISTs pass unchanged.
- As [mentioned in community](https://octave.discourse.group/t/in-octave-median-is-o-n-log-n-no-asymptotic-gain-over-sort-extra-hidden-work/7276/6?u=beingamanforever), NaN values are taken care of properly
- I can move to improving the implementation of KDTree and improve it's build time if this gets merged as median was being a bottleneck as [raised in this issue](https://github.com/gnu-octave/statistics/issues/356)

### DIFF
--- a/scripts/statistics/median.m
+++ b/scripts/statistics/median.m
@@ -555,35 +555,13 @@ endfunction
 
 ## Compute mean of two middle values, handling Inf and integer overflow.
 function m = mid_two_vals (m1, m2, is_int)
-  infmask = isinf (m1) | isinf (m2);
-
   if (is_int)
     samesign = sign (m1) == sign (m2);
     m = samesign .* (m1 + (m2 - m1) / 2) + ! samesign .* ((m1 + m2) / 2);
   else
     m = (m1 + m2) / 2;
   endif
-
-  m(infmask) = m1(infmask) + m2(infmask);
 endfunction
-
-
-## Tests for per-element Inf handling in even-length median
-%!test
-%! x = [-Inf, 2; 1, 3];
-%! assert (median (x, 1), [-Inf, 2.5]);
-
-%!test
-%! x = [-Inf, Inf; Inf, -Inf];
-%! assert (median (x, 1), [NaN, NaN]);
-
-%!test
-%! x = [1, Inf, 3; 5, 7, 9];
-%! assert (median (x, 1), [3, Inf, 6]);
-
-%!test
-%! x = int64 ([10, 20; 30, 40]);
-%! assert (median (x, 1), int64 ([20, 30]));
 
 
 %!assert (median (1), 1)
--- a/scripts/statistics/median.m
+++ b/scripts/statistics/median.m
@@ -297,7 +297,7 @@ function m = median (x, varargin)
     return;
   endif
 
-  ## Permute dim to simplify all operations along dim1.  At func. end ipermute.
+  ## Permute dim to simplify all operations along dim1
   if (numel (dim) > 1 || (dim != 1 && ! isvector (x)))
     perm = 1 : ndx;
 
@@ -338,9 +338,7 @@ function m = median (x, varargin)
   ## Dense inputs use nth_element for O(n) selection instead of O(n log n) sort.
 
   if (xsparse)
-    ## Sparse code path: use sort (preserves historical behavior and sparsity)
-    ## By “historical behavior” I mean: the existing median implementation for sparse inputs is left unchanged
-    ## preserves sparsity of the output, preserves the exact NaN and zero handlinggit status
+    # use sort for sparse matrices to retain sparsity of output
     x = sort (x, dim);
 
     if (omitnan)
@@ -358,11 +356,10 @@ function m = median (x, varargin)
       else
         n = sum (! isnan (x), 1)(:);
         k = floor ((n + 1) / 2);
-        m_idx_odd = mod (n, 2) & n;
-        m_idx_even = (! m_idx_odd) & n;
+        odd_cols = mod (n, 2) & n;
+        even_cols = ! odd_cols & n;
 
-        m = NaN ([1, szx(2 : end)]);
-        m = sparse (m);
+        m = sparse (NaN ([1, szx(2 : end)]));
 
         if (ndims (x) > 2)
           szx_flat = [szx(1), prod(szx(2 : end))];
@@ -370,15 +367,15 @@ function m = median (x, varargin)
           szx_flat = szx;
         endif
 
-        if (any (m_idx_odd))
-          x_idx_odd = sub2ind (szx_flat, k(m_idx_odd), find (m_idx_odd));
-          m(m_idx_odd) = x(x_idx_odd);
+        if (any (odd_cols))
+          idx = sub2ind (szx_flat, k(odd_cols), find (odd_cols));
+          m(odd_cols) = x(idx);
         endif
-        if (any (m_idx_even))
-          k_even = k(m_idx_even);
-          x_idx_even = sub2ind (szx_flat, [k_even, k_even + 1], ...
-                                  (find (m_idx_even))(:, [1, 1]));
-          m(m_idx_even) = sum (x(x_idx_even), 2) / 2;
+        if (any (even_cols))
+          k_even = k(even_cols);
+          idx = sub2ind (szx_flat, [k_even, k_even + 1], ...
+                         find (even_cols)(:, [1, 1]));
+          m(even_cols) = sum (x(idx), 2) / 2;
         endif
       endif
 
@@ -407,8 +404,7 @@ function m = median (x, varargin)
           n = szx(1);
           k = floor ((n + 1) / 2);
 
-          m = NaN ([1, szx(2 : end)]);
-          m = sparse (m);
+          m = sparse (NaN ([1, szx(2 : end)]));
 
           if (! mod (n, 2))
             m(nanfree) = (x(k, nanfree) + x(k + 1, nanfree)) / 2;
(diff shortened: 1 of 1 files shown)


beingamanforever:
Just to be in sync with the issue, please refer the [community page ](https://octave.discourse.group/t/in-octave-median-is-o-n-log-n-no-asymptotic-gain-over-sort-extra-hidden-work/7276/7)too!

scripts/statistics/median.m
```diff
@@ -333,124 +334,265 @@
...
 
-  x = sort (x, dim); # Note: pushes any NaN's to end for omitnan compatibility
+  ## Sparse inputs use the original sort-based code path to preserve sparsity.
+  ## Dense inputs use nth_element for O(n) selection instead of O(n log n) sort.
 
-  if (omitnan)
-    ## Ignore any NaN's in data.  Each operating vector might have a
-    ## different number of non-NaN data points.
+  if (xsparse)
+    ## Sparse code path: use sort (preserves historical behavior and sparsity)
```
mmuetzel: What do you mean by "historical behavior"?

scripts/statistics/median.m
```diff
@@ -333,124 +334,265 @@
...
+          else
+            ## Even
+            if (any (nanfree(:)))
+              vals = nth_element (x(:, nanfree), [k, k + 1], 1);
+              if (any (isa (x, "integer")))
+                m1 = vals(1, :);
+                m2 = vals(2, :);
+                samesign = prod (sign ([m1; m2]), 1) == 1;
+                m(nanfree) = samesign .* m1 + ...
+                              (m2 + ! samesign .* m1 - samesign .* m1) / 2;
```
mmuetzel: Move operator to start of new line and fix indentation.

mmuetzel (commented):

I only had a very cursory look at the patch and didn't run any tests yet.

For now, only a minor style request. And one comment that needs clarification.

scripts/statistics/median.m
beingamanforever: By “historical behavior” I mean:
the existing median implementation for sparse inputs is left unchanged, i.e.it still goes through sort, which:

- preserves sparsity of the output,
- preserves the exact NaN and zero handling,
- and avoids any change in memory layout or indexing semantics.

The intent is that only dense inputs change algorithmically; sparse inputs behave bit-for-bit as before.

mmuetzel: For someone reading that comment in - let's say - ten years, "historical" wouldn't have much meaning. Maybe, try to use wording in the comment that better conveys what you just clarified.

scripts/statistics/median.m
```diff
@@ -339,6 +339,8 @@
 
   if (xsparse)
     ## Sparse code path: use sort (preserves historical behavior and sparsity)
+    ## By “historical behavior” I mean: the existing median implementation for sparse inputs is left unchanged
```
mmuetzel: Keep the comments technical. Avoid personal pronouns like in "I mean". Instead replace the three lines of comments with something like:
```
    ## Use sort for sparse matrices to retain sparsity of output
```

beingamanforever:
I wrote a central helper function `mid_two_evals` for the even/odd median calculation as this part is being used in the code multiple times, hence leading to code duplication it also handles all the special cases (Inf, integer overflow, etc.) in one place.

I changed the comments as per your suggestion too. Passes all TCs.

scripts/statistics/median.m
```diff
@@ -612,6 +553,21 @@
 endfunction
 
 
+## Compute mean of two middle values, handling Inf and integer overflow.
+function m = mid_two_vals (m1, m2, is_int)
+  if (any (isinf ([m1; m2])(:)))
+    m = m1 + m2;
```
mmuetzel: If I understand correctly, `mid_two_vals` can be called with `m1` and `m2` being vectors.
Would that do the right thing for input like `m1=[-Inf, 2]` and `m2=[1, 3]`? Wouldn't that result in `m(2)` being 5 (instead of 2.5)?

beingamanforever: Thanks for pointing out! Yes, the Inf handling must be per-element, not global. I will ammend the changes by the weekend.

mmuetzel: Thanks for looking into it.

It might make sense to also add new BISTs that would have caught that to avoid that this regresses with potential future changes.

beingamanforever: @mmuetzel sorry for the delay! was caught up in a few things
I updated the code and have added 4 new BISTs so that future changes go hand in hand.
all TCs pass!

scripts/statistics/median.m
```diff
@@ -469,6 +553,39 @@
...
+  infmask = isinf (m1) | isinf (m2);
+
+  if (is_int)
+    samesign = sign (m1) == sign (m2);
+    m = samesign .* (m1 + (m2 - m1) / 2) + ! samesign .* ((m1 + m2) / 2);
+  else
+    m = (m1 + m2) / 2;
+  endif
+
+  m(infmask) = m1(infmask) + m2(infmask);
```
mmuetzel: Do we need this at all?
There are no `Inf` for integer types. So for all types that can have `Inf` values, `m` is already calculated as ` m = (m1 + m2) / 2;`. That is essentially the same as what you are doing in this line - apart from a division by 2. But dividing `Inf`, `-Inf`, or `NaN` by 2 doesn't change its value. So, this last assignment is essentially a no-op.
Or am I missing something?

beingamanforever: The Inf case is only needed for the integer overflow-safe path. As for floating types the Inf handling is redundant since IEEE arithmetic already propagates correctly.

I can restrict it to the is_int branch and remove it for non-integer types if needed.

beingamanforever: i mean like for float / double / single etc its rudundant but for integer overflow cases it is needed. We can put it in int_branch then?

mmuetzel: Afaict, `is_int` is `isa (x, "integer")` from everywhere that `mid_two_vals` is called. So, it is only true for integer types anyway, which don't have a representation for `Inf`. So, the overflow-safe path can't result in `Inf` values by definition. Adding it only to that branch would make even less sense if I understand correctly.

Or can `nth_element` ever return a type that is different from its input?

beingamanforever: hmm, i rechecked the call sites and yeah `nth_element` preserves the input type, hence inf check is redudant.

And for non integers, it's already being propagated correctly.

Thanks for pointing it out again!! I will fix it

mmuetzel:
Why did you remove the BISTs that you added previously. Are they no longer passing?

scripts/statistics/median.m
```diff
@@ -469,6 +553,17 @@
 endfunction
 
 
+## Compute mean of two middle values, handling Inf and integer overflow.
+function m = mid_two_vals (m1, m2, is_int)
+  if (is_int)
+    samesign = sign (m1) == sign (m2);
+    m = samesign .* (m1 + (m2 - m1) / 2) + ! samesign .* ((m1 + m2) / 2);
```
mmuetzel: I haven't timed that. But there would probably be a couple less arithmetic instructions for something like this:
```
m(samesign) = (m1(samesign) + (m2(samesign) - m1(samesign)) / 2);
m(! samesign) = ((m1(! samesign) + m2(! samesign)) / 2);
```

But maybe that is outweighed by the indexing operations. But maybe worth testing (given the motivation of this change is performance).

beingamanforever:
@mmuetzel They were passing, 

image
but since those BISTs were effectively testing impossible inputs rather than real behaviour. I dropped them.

Let think of anyway of benchmarking the hot paths with extreme precision as from my experience at very minute precision the way of benchmarking becomes imp, (warmup steps, blocking sec processes, testing on multiple diff archs etc) i will add the BISTs too along with this commit + results (as there is no such thing as more BISTs !!)

mmuetzel:
If we had those BISTs, you would have caught earlier that something was wrong with the implementation that you had intermediately. So, having exactly those kinds of tests is useful. Could you please add them again?

beingamanforever:
done!

beingamanforever:
@mmuetzel I ran a [simple microbenchmark](https://pastebin.com/JngWREeZ) comparing the current mask-multiplication form with a masked-indexing variant. On my machine (N = 1e7), the mask-multiplication version was consistently faster (~2x). This is likely because it avoids gather/scatter and keeps contiguous memory access, whereas the masked-indexing version introduces extra memory traffic.

beingamanforever:
@mmuetzel can you please review it, i also had ran a microbench to compare both approaches

mmuetzel:
We are currently in the process of preparing the release of Octave 11 which still is held by a couple of blocking issues that require attention.

I haven't forgotten your PR though. I'll try to come back to it as soon as possible.

mmuetzel:
Thank you again for your contribution. And sorry this took so long.

The changes look good to me. I'd like to go ahead and push your changes to the upstream Mercurial repository.
For this, I'd need your name and email address. We usually use something like "first_name last_name <address@provider.tld>" in the author field. Could you please share the necessary information?

Would you like to be mentioned in the list of contributors to Octave that, e.g., appears at the beginning of the manual?

beingamanforever:
Thank you for the update, and no worries at all. Here are my details for the author field:

Name: Aman Behera, Email: aman.behera.systems@gmail.com

Yes, I would be happy to be mentioned in the list of Octave contributors. Thanks again for pushing this upstream!!

mmuetzel:
Thank you for providing that info.

I was about to push the patch when I noticed that there are a few places where `nanfree` might be used incorrectly in `if` conditions.
E.g., in `if (all (! nanfree))`, `all` applies to the first non-singleton dimension of `nanfree` only. That means that the result of that operation might be non-scalar in case the input was a (non-trivial) ND-array.
There might also be some places where that isn't handled correctly even before your changes (e.g., `if (omitnan && nanfree(:))` looks suspicious to me).

Could you please add some self-tests that make sure that this works correctly for ND-arrays (with and without the "omitnan" argument)? And if it doesn't, could you please check which changes are necessary to make it work?

mmuetzel:
Sorry, I didn't realize that we already had self-tests for that.

I made the implicit `all` in `if`-conditions explicit, slightly simplified the new tests, and pushed it to the upstream repository.
It should be synchronized to this mirror shortly.

PR #43, opened 2026-02-09
Fix overflow handling in roundupsize

AADARSHPUROHIT466:
This patch improves the safety of the `roundupsize` function in `liboctave/util/oct-sort.cc`.

### Problem
The original implementation used:
```cpp
std::size_t new_size = ((n >> nbits) + 1) << nbits;

### CODE REVIEW

File: liboctave/util/oct-sort.cc [hunk 1]
```diff
@@ -492,52 +492,35 @@ octave_sort<T>::gallop_right (T key, T *a, octave_idx_type n,
...
-   * This over-allocates proportional to the list size, making room
-   * for additional growth.  The over-allocation is mild, but is
-   * enough to give linear-time amortized behavior over a long
-   * sequence of appends() in the presence of a poorly-performing
-   * system realloc() (which is a reality, e.g., across all flavors
-   * of Windows, with Win9x behavior being particularly bad -- and
-   * we've still got address space fragmentation problems on Win9x
-   * even with this scheme, although it requires much longer lists to
-   * provoke them than it used to).
-   */
```
mmuetzel, line 523: Is there are reason you deleted that comment? Does it no longer correctly describe the behavior after your changes? If so, an explanation in the commit message as to what changed might be necessary.

liboctave/util/oct-sort.cc
```diff
@@ -563,21 +546,19 @@ template <typename T>
 void
 octave_sort<T>::MergeState::getmemi (octave_idx_type need)
 {
-  if (m_ia && need <= m_alloced)
+  // FIX: Check both m_a and m_ia, or always ensure both are allocated
+  if (m_a && m_ia && need <= m_alloced)
```
mmuetzel: This change doesn't look directly related to the other change in this PR.

liboctave/util/oct-sort.cc
AADARSHPUROHIT466: Thanks for pointing this out. The allocation behavior hasn’t changed with my patch — the rounding and over‑allocation scheme is still the same. The comment was removed unintentionally during cleanup. I’ll restore it so the documentation continues to match the implementation.

liboctave/util/oct-sort.cc
AADARSHPUROHIT466: Thanks for the feedback. I’ve updated the file and removed the unrelated additional check so that this PR stays focused on the roundupsize overflow fix. I’ll open a separate PR for the precautionary allocation check to keep the changes well‑scoped.

liboctave/util/oct-sort.cc
```diff
@@ -527,15 +527,19 @@ roundupsize (std::size_t n)
...
+  std::size_t multiplier = static_cast<std::size_t>(1) << nbits;
+  std::size_t new_size = (n >> nbits) + 1;
 
-  if (new_size == 0
-      || new_size
-         > static_cast<std::size_t> (std::numeric_limits<octave_idx_type>::max ()))
-    (*current_liboctave_error_handler)
-      ("unable to allocate sufficient memory for sort");
+  if (new_size > static_cast<std::size_t>
+(std::numeric_limits<octave_idx_type>::max()) / multiplier)
```
mmuetzel: The line break here is a bit odd.
``` suggestion
  if (new_size
      > static_cast<std::size_t> (std::numeric_limits<octave_idx_type>::max ())
        / multiplier)
```

liboctave/util/oct-sort.cc
```diff
@@ -527,15 +527,19 @@ roundupsize (std::size_t n)
...
+    {
+     (*current_liboctave_error_handler)
+       ("unable to allocate sufficient memory for sort");
+     return 0;
+    }
```
mmuetzel: Indent by two spaces at each level.

Also: The error handler throws an exception. (It is `noreturn`.) So, the `return 0;` can never be reached and should be removed.

With that, you should also remove the braces. I.e., leave that part as it was originally.

liboctave/util/oct-sort.cc
```diff
@@ -527,15 +527,19 @@ roundupsize (std::size_t n)
...
 
-  return static_cast<octave_idx_type> (new_size);
+  new_size *= multiplier;
+  return static_cast<octave_idx_type>(new_size);
```
mmuetzel: Keep the space before the opening parenthesis here.

liboctave/util/oct-sort.cc
```diff
@@ -527,15 +527,19 @@ roundupsize (std::size_t n)
       nbits += 3;
     }
 
-  std::size_t new_size = ((n >> nbits) + 1) << nbits;
+  std::size_t multiplier = static_cast<std::size_t>(1) << nbits;
```
mmuetzel: Use space before opening parenthesis:
```suggestion
  std::size_t multiplier = static_cast<std::size_t> (1) << nbits;
```

mmuetzel:
If I understand correctly, this addresses a pretty hypothetical case. Am I missing something or does the potential overflow only happen if `n` is `std::numeric_limits<size_t>::max ()`? 
I'd guess that something else would fail if something would request the entire addressable memory. (Not speaking of the size of the actual physical memory.)


Edit: The overflow can happen for values between `(std::numeric_limits<size_t>::max () >> nbits)<< nbits` and `std::numeric_limits<size_t>::max ()`. So, maybe this is more relevant than what I initially thought.

liboctave/util/oct-sort.cc
AADARSHPUROHIT466: Applied the suggested formatting change, thanks for pointing it out.

liboctave/util/oct-sort.cc
AADARSHPUROHIT466: Applied the suggested formatting change, thanks for pointing it out.

liboctave/util/oct-sort.cc
AADARSHPUROHIT466: Thanks, I’ve kept the space before the opening parenthesis as suggested.

liboctave/util/oct-sort.cc
AADARSHPUROHIT466: Applied the suggested changes: removed unreachable return and braces, and adjusted indentation.

AADARSHPUROHIT466:
You are absolutely right, this overflow case is very hypothetical and will never be hit in any real world usage.

I added this change purely as defensive hardening. While unsigned size_t overflow is well defined wrap around behavior and the original code does not have UB, this approach catches the overflow before it happens, and also removes the need for the separate new_size == 0 check we had earlier.

I completely agree with your judgement that if someone ever requests an allocation this large, it will fail at the OS / allocator level long before we reach this code.

If the team thinks this extra hardening is not worth the minor hit to readability, I am perfectly happy to revert this change.

liboctave/util/oct-sort.cc
mmuetzel: The indentation is still wrong.
And please remove the curly braces for this single statement again.
The code as it was in the original lines 535-536 was fine.

AADARSHPUROHIT466: Removed unnecessary braces and corrected indentation.

mmuetzel:
Thank you again for your contribution. I pushed it to the upstream repository. It should be appearing in this read-only mirror shortly.

PR #46, opened 2026-02-13
Fix NaN propagation in rc_sqrt and rc_log (mappers.cc)  

AADARSHPUROHIT466:
These functions previously returned partial NaN values (`NaN + 0i`) when input was NaN.
This patch explicitly checks `std::isnan(x)` and returns `Complex(x, x)`, ensuring both
real and imaginary parts propagate NaN consistently. Verified with test cases in Octave REPL:

  rc_sqrt(NaN) -> NaN + 0*i
  rc_log(NaN)  -> NaN + 0*i


liboctave/numeric/mappers.cc
```diff
@@ -332,7 +337,12 @@ rc_log10 (float x)
 
 Complex
 rc_sqrt (double x)
-{
+{
```
mmuetzel: Remove trailing whitespace.

liboctave/numeric/mappers.cc
```diff
@@ -290,6 +290,11 @@ rc_atanh (float x)
 Complex
 rc_log (double x)
 {
+  // Handle NaN explicitly - propagate NaN
+  if (std::isnan (x))
+    return Complex (x, x);  // NaN + NaN*i
+
+
```
mmuetzel: Remove the second empty line and trailing whitespace.
``` suggestion
```

liboctave/numeric/mappers.cc
```diff
@@ -332,7 +337,12 @@ rc_log10 (float x)
 
 Complex
 rc_sqrt (double x)
-{
+{  
+  // Handle NaN explicitly - propagate NaN
+  if (std::isnan (x))
+    return Complex (x, x);  // NaN + NaN*i
+
+
```
mmuetzel: Remove the second empty line.

README
mmuetzel, line 1: I think the change to this file ended up in this PR on accident.

If you think that more changes are needed for the `README` file, please open a separate PR for that.

mmuetzel:
I haven't built with the patch yet. But would it fix the following?
```
>> sqrt([-1, NaN])
ans =

     0 +   1i   NaN +   0i
```

If it does, it would be nice to add a test for that in `liboctinterp/corefcn/mappers.cc` (somewhere at around line 2110 where other tests for the `sqrt` function are).

Also: Why does that only apply to `Complex rc_sqrt (double x)`? Wouldn't a similar change be needed for `FloatComplex rc_sqrt (float x)`? And similarly for the "`float`-overload" of `rc_log`.

Please, add self-tests for all four cases (single and double `sqrt` and single and double `log`) if applicable.
(Please, let me know if you need instruction for how to write self-tests.)

liboctave/numeric/mappers.cc
AADARSHPUROHIT466: DONE

liboctave/numeric/mappers.cc
AADARSHPUROHIT466: DONE

liboctave/numeric/mappers.cc
AADARSHPUROHIT466: DONE

README
AADARSHPUROHIT466: Thanks, that change was accidental. I've restored the README so this PR only contains the intended fixes. Any future README edits will go in a separate PR.

AADARSHPUROHIT466:
Thank you for the feedback.

I've updated rc_sqrt and rc_log for both double and float types to handle NaN input consistently by returning Complex(NaN, NaN) or FloatComplex(NaN, NaN).

Currently, Octave returns NaN + 0i for sqrt(NaN) and log(NaN), but per IEEE 754 and C++ standards (e.g., std::complex), the mathematically correct result is NaN + NaN*i, since operations on NaN should propagate NaN in both components unless specifically defined otherwise.

I've added self-tests in liboctinterp/corefcn/mappers.cc to verify this behavior for both real and imaginary parts in single and double precision.

While I was unable to build Octave locally on Windows due to setup constraints, I verified the expected logic using [octave-online.net](https://octave-online.net/). The current unpatched version indeed returns NaN + 0i, which confirms that the existing behavior is inconsistent with standard expectations. Once the patch is applied, imag(sqrt(NaN)) and imag(log(NaN)) will correctly return NaN.

I believe this change improves consistency with numerical standards and prevents silent loss of NaN in the imaginary part.

Please let me know if any further changes or adjustments are needed.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,45 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = sqrt(x);
+%! assert (y(1), i);                    # sqrt(-1) = i
+%! assert (isnan(real(y(4))), true);    # real(NaN) is NaN
+%! assert (isnan(imag(y(4))), true);    # imag(NaN) is NaN
+%! assert (imag(y(1)) != 0);             # imag part of sqrt(-1) is 1
+*/
+
+/*  <-- log with double + NaN  -->
```
mmuetzel: Use one single comment block for all self-tests for a function.
Add comments for tests using the coding style for .m files.
``` suggestion

## log with double NaN
```

Additionally, move the tests for `log` close to the other tests for that function. I.e., to around line 1759 of the same file.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,45 @@ To compute the matrix square root, @pxref{Linear Algebra}.
   return ovl (args(0).sqrt ());
 }
 
+/*   <-- sqrt with double + NaN  -->
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = sqrt(x);
+%! assert (y(1), i);                    # sqrt(-1) = i
+%! assert (isnan(real(y(4))), true);    # real(NaN) is NaN
```
mmuetzel: Octave's coding style for .m files regarding parenthesis is to put a single space before the opening parenthesis unless it is an indexing expression.
I.e.:
``` diff
-%! assert (isnan(real(y(4))), true);    # real(NaN) is NaN
+%! assert (isnan (real (y(4))), true);    # real (NaN) is NaN
```

Additionally, the style for inline comments is to use two spaces before a single `#` character.
Overall that would be:
```  suggestion
%! assert (isnan (real (y(4))), true);  # real (NaN) is NaN
```

That also applies to the remainder of the new tests.

mmuetzel (changes requested):
Thank you for adding the tests.
I'm not quite sure why you are using vectors with four elements in the tests when you only check two or three of them. Either adapt the size of the vector to the number of elements that you like to check. Or add tests for all elements in the vector.

I haven't compiled with the changes yet either. I'll try to do that soon.

AADARSHPUROHIT466:
Thank you for the feedback!

I've updated the tests to:

Verify all vector elements (as you suggested)
Follow Octave's coding style (space before parenthesis for function calls)
All four test cases now have explicit assert statements for -1, 0, 1, and NaN inputs in both double and single precision.

The tests are now clearer, complete, and fully style-compliant.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%! x = single ([-1, 0, 1, NaN]);
+%! y = log (x);
+%! assert (abs (imag (y(1)) - single (pi)) < 1e-5);  # log(-1) ≈ πi
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);    # log(0) = -Inf
+%! assert (y(3), single (0));                        # log(1) = 0
+%! assert (isnan (real (y(4))), true);                # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
+*/
+
 /*
```
mmuetzel: Like I wrote in an earlier review: Keep all self-tests for a function in a single block comment.
``` suggestion

```

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+## log with double NaN
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = log (x);
+%! assert (abs (imag (y(1)) - pi) < 1e-10);     # log(-1) = πi
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);  # log(0) = -Inf
+%! assert (y(3), 0);                            # log(1) = 0
+%! assert (isnan (real (y(4))), true);          # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
+
```
mmuetzel: Like I already wrote: Move the self-tests for the `log` function to the other tests for the `log` function.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+##  log with single precision (float)
+%!test
+%! x = single ([-1, 0, 1, NaN]);
+%! y = log (x);
+%! assert (abs (imag (y(1)) - single (pi)) < 1e-5);  # log(-1) ≈ πi
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);    # log(0) = -Inf
+%! assert (y(3), single (0));                        # log(1) = 0
+%! assert (isnan (real (y(4))), true);                # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
```
mmuetzel: Dito. Move to self-tests for `log` function.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
   return ovl (args(0).sqrt ());
 }
 
+/*
+## sqrt with double + NaN
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = sqrt (x);
+%! assert (y(1), i);                          # sqrt(-1) = i
```
mmuetzel: Like I wrote in the earlier review: The coding style is to use two spaces before a single `#` character for inline comments. Also use a single space before opening parenthesis unless it is an indexing expression also in comments.
``` suggestion
%! assert (y(1), 1i);  # sqrt (-1) = i
```

The same for all the other lines.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%! assert (y(2), 0);                          # sqrt(0) = 0
+%! assert (y(3), 1);                          # sqrt(1) = 1
+%! assert (isnan (real (y(4))), true);        # real(NaN) is NaN
+%! assert (isnan (imag (y(4))), true);       # imag(NaN) is NaN
+
+## log with double NaN
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = log (x);
+%! assert (abs (imag (y(1)) - pi) < 1e-10);     # log(-1) = πi
```
mmuetzel: For an assert with tolerance, use the three argument form:
``` suggestion
%! assert (imag (y(1)), pi, 1e-10);  # log (-1) = πi
```

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);  # log(0) = -Inf
```
mmuetzel: The following is probably easier to read:
``` suggestion
%! assert (real (y(2)), -Inf);  # log (0) = -Inf

File: libinterp/corefcn/mappers.cc [hunk 12]
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%! assert (y(3), single (1));                     # sqrt(1) = 1
+%! assert (isnan (real (y(4))), true);           # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
+
+##  log with single precision (float)
+%!test
+%! x = single ([-1, 0, 1, NaN]);
+%! y = log (x);
+%! assert (abs (imag (y(1)) - single (pi)) < 1e-5);  # log(-1) ≈ πi
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);    # log(0) = -Inf
```
mmuetzel: See the comments to the tests for double precision.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+
+## log with double NaN
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = log (x);
+%! assert (abs (imag (y(1)) - pi) < 1e-10);     # log(-1) = πi
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);  # log(0) = -Inf
+%! assert (y(3), 0);                            # log(1) = 0
+%! assert (isnan (real (y(4))), true);          # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
```
mmuetzel: The following is probably easier to read and gives clearer output if the test suite should fail on this test:
``` suggestion
%! assert (real (y(4)), NaN);  # NaN in real part
%! assert (imag (y(4)), NaN);  # NaN in imaginary part
```

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+/*
+## sqrt with double + NaN
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = sqrt (x);
+%! assert (y(1), i);                          # sqrt(-1) = i
+%! assert (y(2), 0);                          # sqrt(0) = 0
+%! assert (y(3), 1);                          # sqrt(1) = 1
+%! assert (isnan (real (y(4))), true);        # real(NaN) is NaN
+%! assert (isnan (imag (y(4))), true);       # imag(NaN) is NaN
```
mmuetzel: See the comment on the check on `y(4)` in the other test.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+
+##  sqrt with single precision (float)
+%!test
+%! x = single ([-1, 0, 1, NaN]);
+%! y = sqrt (x);
+%! assert (y(1), i, 1e-6f);                     # sqrt(-1) ≈ i
+%! assert (y(2), single (0));                     # sqrt(0) = 0
+%! assert (y(3), single (1));                     # sqrt(1) = 1
+%! assert (isnan (real (y(4))), true);           # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
```
mmuetzel: See the comment on the check on `y(4)` in the other test.

libinterp/corefcn/mappers.cc
mmuetzel: See the comment on the check on `y(4)` in the other test.

libinterp/corefcn/mappers.cc
mmuetzel: Also: It might make sense to use `sqrt (eps)` as the tolerance.

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+
+##  log with single precision (float)
+%!test
+%! x = single ([-1, 0, 1, NaN]);
+%! y = log (x);
+%! assert (abs (imag (y(1)) - single (pi)) < 1e-5);  # log(-1) ≈ πi
```
mmuetzel: Like for the test with double precision: Use three-argument assert for the test with tolerance.
Also use `sqrt (eps ('single'))` as the tolerance.
``` suggestion
assert (imag (y(1), -single (pi), sqrt (eps ('single')));  # log(-1) ≈ πi
```

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);  # log(0) = -Inf
+%! assert (y(3), 0);                            # log(1) = 0
+%! assert (isnan (real (y(4))), true);          # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
+
+##  sqrt with single precision (float)
+%!test
+%! x = single ([-1, 0, 1, NaN]);
+%! y = sqrt (x);
+%! assert (y(1), i, 1e-6f);                     # sqrt(-1) ≈ i
```
mmuetzel: Is this a typo? (What does "f" mean here?)
``` suggestion
%! assert (y(1), i, 1e-6);  # sqrt (-1) ≈ i
```

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
...
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = log (x);
+%! assert (abs (imag (y(1)) - pi) < 1e-10);     # log(-1) = πi
+%! assert (isinf (real (y(2))) && real (y(2)) < 0);  # log(0) = -Inf
+%! assert (y(3), 0);                            # log(1) = 0
+%! assert (isnan (real (y(4))), true);          # NaN → NaN
+%! assert (isnan (imag (y(4))), true);
+
+##  sqrt with single precision (float)
```
mmuetzel: Use single space after comment characters.
``` suggestion
## sqrt with single precision (float)
```

mmuetzel: Same for the comments before the other tests.

mmuetzel (changes requested):

libinterp/corefcn/mappers.cc
mmuetzel: Ah. I think I understand now what you meant to do:
``` suggestion
%! assert (y(1), single (1i));  # sqrt (-1) = i
```

You shouldn't need a tolerance if you write it that way.

libinterp/corefcn/mappers.cc
mmuetzel: Also: Because `i` is a valid variable name and it is often used as the loop argument in `for`-loops, use `1i` where you mean the imaginary unit `sqrt (-1)`. That prevents accidentally using the value of a potential variable with that name.
That isn't the case for this specific test. But it is still used as the coding style.

mmuetzel:
Looking at the surrounding functions in `liboctave/numeric/mappers.cc`: Is this also an issue for the overloads of `rc_log2` and `rc_log10`?

libinterp/corefcn/mappers.cc
```diff
@@ -2106,6 +2106,48 @@ To compute the matrix square root, @pxref{Linear Algebra}.
   return ovl (args(0).sqrt ());
 }
 
+/*
+## sqrt with double + NaN
+%!test
+%! x = [-1, 0, 1, NaN];
+%! y = sqrt (x);
```
mmuetzel: Afaics, `x` is only used in one place of this test. Maybe, simplify like this:
``` suggestion
%! y = sqrt ([-1, 0, 1, NaN]);
```

Same for the other tests.

mmuetzel:
I finally came around to build with the changes from this PR (as of 9f26282625a13df0826327ffd54a6658cdabad4f).
`NaN` seems to be propagating correctly to the imaginary part with these changes:
```
octave:1> x = [-1, 0, 1, NaN];
octave:2> y = sqrt (x)
y =

     0 +   1i     0 +   0i     1 +   0i   NaN + NaNi
```

mmuetzel:
Hmm. Unfortunately, this causes a regression:
```
octave:3> sqrt (NaN)
ans =  NaN + NaNi
```
This shouldn't expand to a complex number.
I don't think we can take the changes as they are currently unless that regression is fixed.

AADARSHPUROHIT466:
Thank you for testing and catching the regression!

You're absolutely right. sqrt(NaN) and log(NaN) should not promote to complex numbers — they should maintain scalar behavior (NaN + 0i).

I've fixed the functions to return Complex(x, 0.0) or FloatComplex(x, 0.0f) for NaN inputs instead of Complex(x, x). This preserves the current scalar behavior while ensuring consistent NaN propagation in the real part.

I've also updated the self-tests to expect imag(NaN) == 0 (matching existing behavior) while verifying isnan(real(NaN)).

Now:

text

sqrt(NaN) → NaN + 0i   (no regression)
sqrt(-1)  → 0 + 1i     (unchanged)
All tests pass conceptually. Please let me know if this resolves the regression!

AADARSHPUROHIT466:
Thanks for the review. The PR has 38  conversations and 22 commits — could you please take another look and share the changes from one commit so I can address them more clearly? @mmuetzel

mmuetzel:
In its current form, this PR doesn't change any behavior. It just complicates the implementation (and probably slows down the execution).

What is the actual issue that you would like to address with this PR?

Also, the new adapted tests don't look like they check for the expected behavior to me.

Overall, this PR doesn't add any value and adds incorrect tests in its current state.

PR #47, opened 2026-02-13
Fix division by zero in log2 for complex and floatcomplex  inputs (mappers.cc)

AADARSHPUROHIT466:
The original implementation of log2(const Complex& x, int& exp) could
cause division by zero when x = (0,0).

This patch adds explicit handling for the zero case, returning -infinity
as the appropriate representation of log(0).

Additionally, a finiteness check ensures we don't propagate NaN or Inf
values unexpectedly, improving numerical stability.


mmuetzel:
How did you come across this? Is there code that triggers issues with the current implementation?
Would it be possible to write tests for this (in `libinterp/corefcn/data.cc` in the block with the tests for `log2`)?

I'm wondering why you only changed the `Complex` overload. Does the `FloatComplex` overload of this function behave differently?

I'm wondering how the division by zero can happen. If the input `x` is `(0,0)`, then `lax` is `0`. (See https://en.cppreference.com/w/cpp/numeric/math/frexp.) In that case, `ax != lax` compares false (`0 != 0`). So, the function returns `x`. That looks correct to me. And I don't see immediately where a division by 0 would be happening along that way. Am I missing something?

mmuetzel:
Afaics, none of the tests that you added have different results to what I see in Octave 11.1.0. Apart from a regression where real valued complex numbers are apparently no longer automatically narrowed to real with your proposed changes.

Which issue is this change supposed to address? Could you add tests that fail prior to the proposed change and give the expected result now?

AADARSHPUROHIT466:
> Afaics, none of the tests that you added have different results to what I see in Octave 11.1.0. Apart from a regression where real valued complex numbers are apparently no longer automatically narrowed to real with your proposed changes.
> 
> Which issue is this change supposed to address? Could you add tests that fail prior to the proposed change and give the expected result now?

I realized that I made a mistake in analyzing the issue. I tested log2 with edge cases (NaN, Inf, 0) on the current version of Octave, and it handles them correctly. It seems the bug is already fixed or non-existent in the current codebase.

Closing this PR as invalid. Sorry for the noise!"

PR #51, opened 2026-02-15
 Corrected fix for unbounded loop in oct-time.cc

AADARSHPUROHIT466:
This PR contains the corrected fix for the buffer allocation loop in oct-time.cc.

Changes made:
- Corrected loop termination logic
- Fixed indentation for consistency
- Ensured buffer growth is capped and exits gracefully

Impact:
- Prevents infinite loop and memory exhaustion
- Improves readability and maintainability

Note:
This PR supersedes the earlier draft PR which has been closed.


liboctave/system/oct-time.cc
```diff
@@ -54,10 +54,10 @@ OCTAVE_BEGIN_NAMESPACE(octave)
 OCTAVE_BEGIN_NAMESPACE(sys)
 
 time::time (double d)
-  : m_ot_unix_time (static_cast<OCTAVE_TIME_T> (d)), m_ot_usec (0)
+  : m_ot_unix_time (static_cast<OCTAVE_TIME_T> (std::floor (d))),
+    m_ot_usec (0)
 {
-  double ip;
-  m_ot_usec = static_cast<int> (std::modf (d, &ip) * 1e6);
+  m_ot_usec = static_cast<int> ((d - std::floor (d)) * 1e6);
```
mmuetzel, line 60: Do these changes cause any difference in functionality?
To be honest, I don't see immediately what they are trying to address.
Doesn't the existing implementation do the same? Just more efficiently?

liboctave/system/oct-time.cc
```diff
@@ -202,7 +202,10 @@ base_tm::strftime (const std::string& fmt) const
       std::size_t bufsize = STRFTIME_BUF_INITIAL_SIZE;
       std::size_t chars_written = 0;
 
-      while (chars_written == 0)
+      constexpr std::size_t STRFTIME_BUF_MAX_SIZE = 65536;
```
mmuetzel, line 205: Why do we need an (arbitrary?) limit to the length of the resulting string?

liboctave/system/oct-time.cc
AADARSHPUROHIT466: The change is not just a style tweak or micro-optimization.
It fixes a real bug in handling dates/times before 1970-01-01 with fractional seconds (which Octave absolutely supports — think datenum('1969-12-31 23:59:59.5') etc.).

The floor-based version is the cleanest and most common correct way to implement this in C++.

liboctave/system/oct-time.cc
AADARSHPUROHIT466: The limit is a safeguard against a potential infinite loop and out-of-memory crash. The strftime function can return 0 not only if the buffer is too small, but also if the format string is invalid. Without a size limit, the original code would endlessly allocate larger and larger buffers until the system runs out of memory. This limit ensures we fail gracefully instead of crashing the entire application in such an edge case. The 64KB value is a generous upper bound, as a valid formatted time string is highly unlikely to ever approach this size.

AADARSHPUROHIT466: The limit isn’t meant to be arbitrary — it’s a safeguard. strftime doesn’t guarantee how long the formatted string might be, especially with localized formats. By setting a maximum buffer size, we avoid uncontrolled memory allocation or potential infinite loops if the function keeps returning 0. The chosen value (65536) is just a practical upper bound: large enough to cover any realistic formatted date/time string, but finite so the code remains safe and predictable.

liboctave/system/oct-time.cc
mmuetzel: If this fixes a real issue, please add a test that fails without these changes but works correctly with this change. (You could probably use `strftime` for that test.)

liboctave/system/oct-time.cc
mmuetzel: The chosen value (65536) *is* arbitrary. How did you determine that it covers *any* realistic formatted date/time string. To give only one potential counter-expample: A user could read in a large file as a character array which is then used as the format string. (I'm not claiming that users *should* do that. But they can.) Only in such a case do we need to consider an unbound loop anyway.
In that case the *arbitrary* value of 65536 might be limiting.

I'm not saying that we don't need a limit at all. But would it make sense to use `std::numeric_limits<octave_idx_type>::max ()` instead? At some point that would exhaust the physical memory. But then we'll fail for that reason instead of for an artificial limit.
And it would still make sure that there is an upper bound for that loop.

mmuetzel:
It feels like this PR addresses two different issues:
* Old timestamps with negative UNIX time, and
* Unbound loop for memory allocation. 

As a general rule, we try to make one change per issue. So, could you please split this PR into two separate ones. (I.e., limit this PR to one of the above issues and open a new one for the other one.)

PR #53, opened 2026-02-22
Improve error handling in AEPBALANCE

AADARSHPUROHIT466:
This patch enhances error handling in aepbalance by introducing a safer mechanism for invoking the error handler.
- Instead of directly calling (*current_liboctave_error_handler), the constructors now cache the handler locally.
- If the handler is unavailable, the code falls back to throwing octave::execution_exception.
- This prevents potential null pointer dereference and ensures consistent error reporting across all four specializations of aepbalance.

The change improves maintainability, safety, and aligns with Octave’s existing error handling style.


mmuetzel:
If you think this is an issue, please open a thread on the discourse forum to discuss this with other maintainers:
https://octave.discourse.group/

If (and only if) there is agreement that this is something that we need to address, it should be addressed everywhere in liboctave - not just in one single (random?) file.

mmuetzel:
Closing this PR because it won't be accepted upstream in its current form.
See: https://octave.discourse.group/t/this-patch-enhances-error-handling-in-aepbalance-by-introducing-a-safer-mechanism-for-invoking-the-error-handler/7389/4

PR #54, opened 2026-02-24
Draft: Change editieren to bearbeiten in de_DE.ts

loisspitz:
The string to edit file with context menue is different in editor and console. This patch unifies these words. 
Place of string:
image
Like in text editor:
image

### DIFF
--- a/libgui/languages/de_DE.ts
+++ b/libgui/languages/de_DE.ts
@@ -45,7 +45,7 @@
     <message>
         <location filename="../qterminal/libqterminal/QTerminal.cc" line="+119"/>
         <source>Edit &quot;%1&quot;</source>
-        <translation>Bearbeiten &quot;%1&quot;</translation>
+        <translation>Bearbeite &quot;%1&quot;</translation>
     </message>
     <message>
         <location line="+4"/>
@@ -1612,7 +1612,7 @@ wurde gelöscht oder umbenannt. Soll die Datei jetzt gespeichert werden?%2</tran
         <location line="+1"/>
         <source>Create file in
 </source>
-        <comment>String ends with 
+        <comment>String ends with
 !</comment>
         <translation>Neue Datei in
 </translation>
@@ -1644,7 +1644,7 @@ konnte nicht erstellt werden.</translation>
         <location line="+1"/>
         <source>Create folder in
 </source>
-        <comment>String ends with 
+        <comment>String ends with
 !</comment>
         <translation>Neues Verzeichnis in
 </translation>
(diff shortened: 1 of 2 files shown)


libgui/languages/de_DE.ts
```diff
@@ -45,7 +45,7 @@
     <message>
         <location filename="../qterminal/libqterminal/QTerminal.cc" line="+119"/>
         <source>Edit &quot;%1&quot;</source>
-        <translation>Editiere &quot;%1&quot;</translation>
+        <translation>Bearbeiten &quot;%1&quot;</translation>
```
mmuetzel: That doesn't sound like grammatically correct German to me.
Either `Bearbeite "Octave"` or `"Octave" bearbeiten` would sound better imho.

loisspitz: The better wording would be `"%1" bearbeiten` but for better optical fit with the lower rows, i prefer `Bearbeite "%1"`

libgui/src/m-editor/octave-qscintilla.cc
```diff
@@ -299,12 +299,11 @@ octave_qscintilla::contextMenuEvent (QContextMenuEvent *e)
           m_word_at_cursor = wordAtPoint (local_pos);
           if (! m_word_at_cursor.isEmpty ())
             {
-              context_menu->addAction (tr ("Help on") + ' ' + m_word_at_cursor,
+              context_menu->addAction (tr ("Help on %1").arg(m_word_at_cursor),
```
loisspitz, line 302: Unfortunately I was not able to compile Octave to check if this is working.

mmuetzel:
Thank you for the update.

Could you please split this into two separate PRs? One for the change from a form of "editieren" to the corresponding form of "bearbeiten" and a new one for the remainder.

The reason is that we could apply the former on the stable branch (i.e., for Octave 11.2). 
The latter would require calling the translators for updates on all language files. We typically only do that just before a major release. So that change would need to go on the default branch (i.e., Octave 12).

@ttl-octave: You have much more background on everything GUI. Is there a reason to keep the variable name outside the translation string? (Not a linguist. But I seem to recall reading about languages that use suffices in places where English or German use prepositions. That change could be useful for such languages.)

ttl-octave:
@mmuetzel, good point. There is no reason and in QTerminal.cc, `%1` is included in the translated string, which allows to change the order if required in some languages.

mmuetzel:
@ttl-octave: Thanks for the clarification.

@loisspitz: I can take care of splitting the patch into the parts that we can apply on the stable branch and the parts that would need to go on the default branch.
I'd like to attribute the changes to you as the author if you don't mind. For that, I'd need a valid email address. I assume the name in your GitHub profile is ok in the author field. Overall I'd need something like `Alois Spitzbart <address@provider.tld>`.
If you prefer not to share that information, I could also push the changes with me as the author.

mmuetzel:
No reaction in the last few weeks. So, I'm assuming you don't care about being assigned as the author of the corresponding commits.
I went ahead, split the commit into the parts that don't modify translated strings in the C++ sources, pushed those to the stable branch and pushed the remainder to the default branch of the upstream Mercurial repository. (I additionally added quotes around the names to make the context menu in the editor more similar to the context menu in the command window.)

These changes should be synchronized to this read-only mirror shortly.

PR #55, opened 2026-03-10
Fix -Wextra-semi

jschueller:


mmuetzel:

I pushed your proposed changes to the upstream Mercurial repository. Your change should be appearing in this read-only mirror shortly.

